PR #8, opened 2023-02-08
Initial resouce connections work

bobbyiliev:
Initial work for the resource connections, eg:

```
resource "materialize_connection" "example_ssh_connection" {
  name            = "example"
  schema_name     = "public"
  connection_type = "SSH TUNNEL"
  ssh_host = "example.com"
  ssh_port = 22
  ssh_user = "example"
}
```

### DIFF
--- a/materialize/resources/config.go
+++ b/materialize/resources/config.go
@@ -3,7 +3,6 @@ package resources
 var connectionTypes = []string{
 	"KAFKA",
 	"POSTGRES",
-	"LOAD GENERATOR",
 	"AWS PRIVATELINK",
 	"SSH TUNNEL",
 	"CONFLUENT SCHEMA REGISTRY",
@@ -44,6 +43,12 @@ var saslMechanisms = []string{
 	"PLAIN",
 	"SCRAM-SHA-256",
 	"SCRAM-SHA-512",
+}
+
+var sourceConnectionTypes = []string{
+	"KAFKA",
+	"POSTGRES",
+	"LOAD GENERATOR",
 }
 
 var sourceSizes = []string{
--- a/materialize/resources/config.go
+++ b/materialize/resources/config.go
@@ -4,6 +4,9 @@ var connectionTypes = []string{
 	"KAFKA",
 	"POSTGRES",
 	"LOAD GENERATOR",
+	"AWS PRIVATELINK"
+	"SSH TUNNEL",
+	"CONFLUENT SCHEMA REGISTRY",
 }
 
 var envelopes = []string{
--- a/materialize/resources/config.go
+++ b/materialize/resources/config.go
@@ -4,7 +4,7 @@ var connectionTypes = []string{
 	"KAFKA",
 	"POSTGRES",
 	"LOAD GENERATOR",
-	"AWS PRIVATELINK"
+	"AWS PRIVATELINK",
 	"SSH TUNNEL",
 	"CONFLUENT SCHEMA REGISTRY",
 }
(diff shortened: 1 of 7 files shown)


materialize/resources/config.go
```diff
@@ -4,6 +4,9 @@ var connectionTypes = []string{
 	"KAFKA",
 	"POSTGRES",
 	"LOAD GENERATOR",
+	"AWS PRIVATELINK",
```
dehume, line 7: Maybe this should get broken out. Since it's used by `SOURCES` this would make these new options allowable for source configurations.

materialize/resources/resource_connection.go
```diff
@@ -0,0 +1,838 @@
...
+				Description: "The name of the connection.",
+				Type:        schema.TypeString,
+				Required:    true,
+			},
+			"schema_name": {
+				Description: "The identifier for the connection schema.",
+				Type:        schema.TypeString,
+				Optional:    true,
+				Default:     "public",
+			},
```
dehume, line 34: Just merged in but I would add a `database_name` value to fully qualify the resource.

materialize/resources/resource_connection.go
```diff
@@ -0,0 +1,838 @@
...
+		FROM mz_connections
+		JOIN mz_schemas
+			ON mz_connections.schema_id = mz_schemas.id
+		WHERE mz_connections.name = '%s'
+		AND mz_schemas.name = '%s';
+	`, b.connectionName, b.schemaName)
+}
+
+func (b *ConnectionBuilder) Rename(newConnectionName string) string {
+	q := strings.Builder{}
```
dehume, line 628: Nit: These simple string parsing you don't have to use `strings.Builder{}` Could probably just be
```
return fmt.Sprintf(`ALTER CONNECTION %s.%s RENAME TO %s.%s;`, b.schemaName, b.connectionName, b.schemaName, newConnectionName)
```

materialize/resources/resource_connection.go
```diff
@@ -0,0 +1,838 @@
...
+				Description: "The SSL mode for the Postgres database.",
+				Type:        schema.TypeString,
+				Optional:    true,
+			},
+			"postgres_aws_privatelink": {
+				Description: "The AWS PrivateLink configuration for the Postgres database.",
+				Type:        schema.TypeString,
+				Optional:    true,
+			},
+			"kafka_broker": {
```
dehume, line 161: Should we remove this and just use `kafka_brokers` and allow 1 or more?

materialize/resources/resource_connection.go
```diff
@@ -0,0 +1,838 @@
...
+		}
+
+		q.WriteString(fmt.Sprintf(`, DATABASE '%s'`, b.postgresDatabase))
+	}
+
+	if b.connectionType == "KAFKA" {
+		if b.kafkaBroker != "" {
+			q.WriteString(fmt.Sprintf(`BROKER '%s',`, b.kafkaBroker))
+		}
+		if b.kafkaBrokers != nil {
```
dehume, line 545: Does this work? I have been checking lists by checking if they are empty or not

bobbyiliev: Just updated this to use the same approach as with tables

materialize/resources/resource_connection.go
bobbyiliev: I like that! Just updated it!

materialize/resources/config.go
bobbyiliev: Just changed that, let me know what you think!

materialize/resources/resource_connection.go
bobbyiliev: As we discussed the just discussed, this follows the [docs](https://materialize.com/docs/sql/create-connection/#general-options), we can later decide if we want to keep it or not.

dehume (approved):

PR #12, opened 2023-02-09
Current Database/Cluster Datasources

dehume:
Add datasource to get the current active database and active cluster. I'm guessing active cluster won't be used as often as active database.


examples/main.tf
```diff
@@ -53,7 +54,8 @@ resource "materialize_source" "example_source_load_generator" {
 
 # Create a secret
 resource "materialize_secret" "example_secret" {
-  name        = "example"
-  schema_name = materialize_schema.example_schema.name
-  value       = "decode('c2VjcmV0Cg==', 'base64')"
+  name          = "example"
+  schema_name   = materialize_schema.example_schema.name
+  database_name = data.materialize_current_database.database.name
+  value         = "decode('c2VjcmV0Cg==', 'base64')"
```
benesch, line 60: Totally unrelated to this PR, but this seems super weird! I think you want the value to be a string that gets escaped and provided directly like `CREATE SECRET s AS '<escaped value goes here>'`.

bobbyiliev: I believe that this is just an example, it will work both ways.

My guess is that it just re-uses the example for the [docs](https://materialize.com/docs/sql/create-secret/)

image

benesch: Right, my point is that this _shouldn't_ work! The Terraform provider should be SQL escaping the provided values, so that if users have e.g. a stray single quote in their secret value, they don't need to escape it themselves!

And then, if this example wanted to show something more complicated, we could rig that up using _Terraform's_ [`base64decode`](https://developer.hashicorp.com/terraform/language/functions/base64decode) function.

Does that make sense? Short on time to explain.

bobbyiliev: Ah I see, yes that makes sense! I've drafted a PR with that change [here](https://github.com/MaterializeInc/terraform-provider-materialize/pull/18)

bobbyiliev (approved):
Looks good! If it's ok we could fix the merge conflicts and merge this, and then sort out the secrets change request in here #18

PR #14, opened 2023-02-10
SQL Error Handling

dehume:
Now SQL related errors will show in the output of the Terraform error. Also the query being executed will be logged at the error level.

### DIFF
--- a/materialize/resources/resource_cluster_replica.go
+++ b/materialize/resources/resource_cluster_replica.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"database/sql"
 	"fmt"
+	"log"
 	"strings"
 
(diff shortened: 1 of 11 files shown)


materialize/resources/resource_cluster_replica.go
```diff
@@ -80,7 +81,7 @@ type ClusterReplicaBuilder struct {
 	idleArrangementMergeEffort int
 }
 
-func newClusterReplicaBuilder(clusterName, replicaName string) *ClusterReplicaBuilder {
+func newClusterReplicaBuilder(replicaName, clusterName string) *ClusterReplicaBuilder {
```
dehume, line 84: Unrelated but noticed cluster replicas were not behaving as expected.

bobbyiliev (approved):
This is great! It's going to make debugging so much easier!

PR #16, opened 2023-02-10
State Import

dehume:
Support for resource imports. The problem with our existing resources is that the id is based on the id from the mz tables. So the `ReadContext` query should be based around filtering on the id. The `CreateContext` will only have access to that id after the resource is created so it will need to run two queries, one to create the resource and one to query the newly created resource to get the id (this will need to change for all our resources).

_main.tf_
```
resource "materialize_secret" "test" {
  name          = "pg_auto_pass"
  schema_name   = "public"
  database_name = "materialize"
}
```
Import resources:
`terraform import materialize_secret.test u2273`

### DIFF
--- a/materialize/resources/resource_secret.go
+++ b/materialize/resources/resource_secret.go
@@ -111,12 +111,11 @@ func readSecretParams(id string) string {
 			ON mz_secrets.schema_id = mz_schemas.id
 		JOIN mz_databases
 			ON mz_schemas.database_id = mz_databases.id
-		WHERE mz_secrets.id = '%s';
-	`, id)
+		WHERE mz_secrets.id = '%s';`, id)
 }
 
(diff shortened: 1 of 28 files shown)


materialize/resources/resource_secret.go
```diff
@@ -40,7 +44,7 @@ func Secret() *schema.Resource {
 			"value": {
 				Description: "The value for the secret. The value expression may not reference any relations, and must be implicitly castable to bytea.",
 				Type:        schema.TypeString,
-				Required:    true,
+				Optional:    true,
```
dehume, line 47: Needed to make value optional, otherwise you can't import an existing secret without setting it which may not be known

dehume (commented):
If this makes sense, can work to implement for the other resources

bobbyiliev (approved):
Just finished testing this and it looks good!

PR #18, opened 2023-02-14
Update secret creation value

bobbyiliev:
As discussed [here](https://github.com/MaterializeInc/terraform-provider-materialize/pull/12/files#r1101929562),  we want the value to be a string that gets escaped and provided directly like `CREATE SECRET s AS '<escaped value goes here>'`.

### DIFF
--- a/materialize/resources/resource_secret.go
+++ b/materialize/resources/resource_secret.go
@@ -95,7 +95,8 @@ func (b *SecretBuilder) Rename(newName string) string {
 }
 
 func (b *SecretBuilder) UpdateValue(newValue string) string {
-	return fmt.Sprintf(`ALTER SECRET %s.%s.%s AS '%s';`, b.databaseName, b.schemaName, b.secretName, newValue)
+	escapedValue := pq.QuoteLiteral(newValue)
+	return fmt.Sprintf(`ALTER SECRET %s.%s.%s AS %s;`, b.databaseName, b.schemaName, b.secretName, escapedValue)
 }
 
 func (b *SecretBuilder) Drop() string {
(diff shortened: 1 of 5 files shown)


materialize/resources/resource_secret.go
```diff
@@ -71,7 +71,7 @@ func newSecretBuilder(secretName, schemaName, databaseName string) *SecretBuilde
 }
 
 func (b *SecretBuilder) Create(value string) string {
-	return fmt.Sprintf(`CREATE SECRET %s.%s.%s AS %s;`, b.databaseName, b.schemaName, b.secretName, value)
+	return fmt.Sprintf(`CREATE SECRET %s.%s.%s AS '%s';`, b.databaseName, b.schemaName, b.secretName, value)
```
benesch, line 74: Eep, this isn't safe! It will fail for secret values that include single quotes. 

The safe way to do this is to use the `pq.QuoteLiteral` function:

https://github.com/lib/pq/blob/d5affd5073b06f745459768de35356df2e5fd91d/conn.go#L1631-L1681

bobbyiliev: Good catch! I've updated this and added a couple more tests for this as well

materialize/resources/resource_secret.go
```diff
@@ -93,7 +95,7 @@ func (b *SecretBuilder) Rename(newName string) string {
 }
 
 func (b *SecretBuilder) UpdateValue(newValue string) string {
-	return fmt.Sprintf(`ALTER SECRET %s.%s.%s AS %s;`, b.databaseName, b.schemaName, b.secretName, newValue)
+	return fmt.Sprintf(`ALTER SECRET %s.%s.%s AS '%s';`, b.databaseName, b.schemaName, b.secretName, newValue)
```
benesch, line 98: This one needs escaping too, yeah?

bobbyiliev: You are too quick :D I pushed this a minute after the first commit!

benesch (approved):
LGTM!

dehume (approved):

PR #20, opened 2023-02-14
Integration Base Testing

dehume:
Adding integration tests for the terraform provider using the `materialized`. This creates a simple docker compose of the `materialized` and a container with the provider installed locally. It then keeps the provider container running (don't love) and then runs a series of individual terraform commands for specific resources using `-target`.


dehume:
@bobbyiliev This works locally. Trying to figure out what the Terraform error could be.

docker-compose.yml
```diff
@@ -0,0 +1,20 @@
...
+version: "3.9"
+
+services:
+
+  materialized:
+    image: materialize/materialized:latest
+    container_name: materialized
+    ports:
+      - 6875:6875
+    healthcheck: {test: curl -f localhost:6875, interval: 1s, start_period: 30s}
```
bobbyiliev, line 11: I believe that the problem is with the healthcheck.
Basically it takes ~10s or so for the materialized container to be fully ready, but only using the `depends_on` directive, would wait for just a second or so.



```suggestion
      - 6878:6878
    healthcheck: {test: curl -f localhost:6878/api/readyz, interval: 1s, start_period: 30s}
```

docker-compose.yml
```diff
@@ -0,0 +1,20 @@
...
+
+  provider:
+    build:
+      context: .
+    container_name: provider
+    depends_on:
+      - materialized
```
bobbyiliev, line 18:
```suggestion
      materialized: {condition: service_healthy}
 ```

docker-compose.yml
bobbyiliev: I've managed to reproduce the failure locally, after running the test a few times in a row until I hit the same error. My guess is that locally the materialized container becomes healthy quicker so we don't hit the problem as often.

bobbyiliev (changes requested):

dehume:
@philip-stoev wanted to get your thoughts on the layout of our integration tests for the Terraform provider. Part of me was thinking we could have individual Terraform commands to spin and down various resources but I could also be convinced that it would be better to just maintain a single stack and spin up and down everything at once. If there are any patterns we use elsewhere, I'm happy to follow those.

dehume:
Setting up integration tests for certain issues and starting to uncover some issues with the more complicated resources. Moving those to separate issues:
* #22 
* #23 
* #24
We may want to add additional services to the docker compose (kafka, postgres...) to handle these. Looking at localstack for AWS Privatelink but there is no localstack option for that.

benesch:
> Looking at localstack for AWS Privatelink but there is no localstack option for that.

That's okay! You can still _create_ the connection in Materialize, with a dummy service name. It just won't connect to anything. That should be sufficient for integration testing the provider. We have a separate end to end test elsewhere that will make sure that AWS PrivateLink connections _actually_ work end to end with real AWS VPC endpoints, etc.

bobbyiliev (approved):

philip-stoev:
@dehume I have zero experience with Terraform, but will try to provide feedback by tomorrow. Please feel free to not wait for me though.

Issue #27, opened 2023-02-18
Escape all identifiers and values


benesch:
## Problem

As far as I can tell, the provider today does not do any escaping of identifiers or values. User-provided values are simply splatted directly into the SQL queries:

https://github.com/MaterializeInc/terraform-provider-materialize/blob/6eaca94205bd2190dd5795c87d5d8f0e77ed2fd7/materialize/resources/resource_cluster.go#L50

https://github.com/MaterializeInc/terraform-provider-materialize/blob/6eaca94205bd2190dd5795c87d5d8f0e77ed2fd7/materialize/resources/resource_cluster_replica.go#L124

I didn't actually try this, but I suspect that this will lead to a syntax error, because `foo@bar` is not a valid identifier:

```hcl
resource "materialize_cluster" "example_cluster" {
  name = "foo@bar"
}
```

And I suspect you can construct classic SQL injection:

```hcl
resource "materialize_cluster" "example_cluster" {
  name = "\"; DROP DATABASE lol; --"
}
```

(SQL injection in this context isn't _really_ a security issue, because you're almost surely running Terraform on trusted input, but it's still a bad look!)

## Solution

Use the following two functions to escape all identifiers and strings before including them in SQL queries:

  * [`pgx.QuoteString`](https://pkg.go.dev/gopkg.in/jackc/pgx.v1#Connection.QuoteString)
  * [`pgx.QuoteIdentifier`](https://pkg.go.dev/gopkg.in/jackc/pgx.v1#Connection.QuoteIdentifier)

This is roughly equivalent to how the Snowflake Terraform provider works.

We should ensure that the tests for all resources have a test case with adversarial values for all parameters.

benesch:
See also #28; the issue description assumes we're going to switch to pgx by recommending the pgx quoting functions. But lib/pq has equivalents, if for some reason we don't want to switch to pgx.

bobbyiliev:
Just had a chat with @dehume about the [pgx.QuoteString](https://pkg.go.dev/gopkg.in/jackc/pgx.v1#Connection.QuoteString) and [pgx.QuoteIdentifier](https://pkg.go.dev/gopkg.in/jackc/pgx.v1#Connection.QuoteIdentifier) functions, as it like those aren't public [functions](https://github.com/jackc/pgx/issues/868), we can't use them directly in our statement builder logic.

Instead, what we could do is to cherry-pick them as both functions are very very simple, just a single line of string replace: 
[...]

PR #32, opened 2023-02-21
Use pgx

dehume:
Use pgx instead of lib/pq for the connection.


materialize/provider.go
```diff
@@ -10,8 +10,8 @@ import (
 
 	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	_ "github.com/jackc/pgx/stdlib"
 	"github.com/jmoiron/sqlx"
-	_ "github.com/lib/pq" //PostgreSQL db
```
benesch, line 14: I think this is getting pulled in somewhere else too, because it's not showing up as removed in go.mod.

dehume: Looks like it is being used for `pq.QuoteLiteral` in the secrets. Looks like this was the suggested solution in https://github.com/MaterializeInc/terraform-provider-materialize/pull/18. Is there a better alternative that doesn't use pg?

benesch: Yes! I linked the equivalent pgx functions here: https://github.com/MaterializeInc/terraform-provider-materialize/issues/27#issue-1590190068

bobbyiliev (approved):
This looks good to me! Leaving it up to @benesch to confirm if he has any other concerns.

benesch (commented):
This is great, thanks! (Sorry, I've not been fastidious with keeping track of PRs in this repo—feel free to ping me on Slack if you need my eyes on something!)

PR #33, opened 2023-02-22
Source Redesign

dehume:
Per @benesch's suggestion breaking up sources into three resources: `materialize_source_kafka`, `materialize_source_loadgen`, `materialize_source_postgres`. Agree that this does make them cleaner and less of a maze to determine which parameters need to be set.

As of now, just leaving the datasource for all sources together. Open to thoughts if we should also break that up.


dehume:
@morsapaes for visibility

bobbyiliev (changes requested):
This looks good I believe! We could use it as a blueprint for the other resources as well.

There seems to be a minor merge conflict

bobbyiliev (approved):
This looks good to me!

PR #37, opened 2023-02-22
Add integration connections tests

### LINKED ISSUE #24: Integration Tests - Connections
Complete integration tests for connections:

- AWS Privatelink
- Schema Registry
- Kafka/Redpanda
- Postgres
- SSH Tunnel

bobbyiliev:
Adding:
 - A new Postgres container, later on could be used for the Postgres source tests
 - SSH connection test
 - Postgres connection test
 - Kafka connection for multiple brokers + SASL test
 
Unfortunately, we can't test PrivateLink connections due to the following blocker:

> Error in mzcompose: AWS PrivateLink connections are not supported 

https://github.com/MaterializeInc/materialize/blob/main/test/testdrive/connection-create-drop.td#L382-L388

This seems to cause TF to panic.

Closes #24

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -13,10 +13,6 @@ jobs:
       - name: Docker Compose Up
         run: docker compose up --build -d
 
-      - name: Configure Materialize
-        run: |
-          docker exec materialized psql -U mz_system -h localhost -p 6877 -d materialize -c 'ALTER SYSTEM SET max_aws_privatelink_connections = 3;'
-
       - name: Terraform Init
         run: |
           docker exec provider terraform init
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -13,6 +13,10 @@ jobs:
       - name: Docker Compose Up
         run: docker compose up --build -d
 
+      - name: Configure Materialize
+        run: |
+          docker exec materialized psql -U mz_system -h localhost -p 6877 -d materialize -c 'ALTER SYSTEM SET max_aws_privatelink_connections = 3;'
+
       - name: Terraform Init
         run: |
           docker exec provider terraform init
@@ -20,7 +24,7 @@ jobs:
       - name: Terraform Apply
         run: |
           docker exec provider terraform apply -auto-approve -compact-warnings
-      
+
       - name: Terraform Destroy
         run: |
           docker exec provider terraform destroy -auto-approve -compact-warnings
--- a/integration/connection.tf
+++ b/integration/connection.tf
@@ -17,3 +17,39 @@ resource "materialize_connection" "schema_registry" {
   connection_type               = "CONFLUENT SCHEMA REGISTRY"
   confluent_schema_registry_url = "http://schema-registry:8081"
 }
+
+resource "materialize_connection" "example_ssh_connection" {
+  name            = "ssh_example_connection"
+  schema_name     = "public"
+  connection_type = "SSH TUNNEL"
+  ssh_host        = "ssh_host"
+  ssh_user        = "ssh_user"
+  ssh_port        = 22
+}
+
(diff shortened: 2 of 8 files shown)


.github/workflows/integration.yml
```diff
@@ -13,17 +13,21 @@ jobs:
       - name: Docker Compose Up
         run: docker compose up --build -d
 
+      - name: Configure Materialize
+        run: |
+          docker exec materialized psql -U mz_system -h localhost -p 6877 -d materialize -c 'ALTER SYSTEM SET max_aws_privatelink_connections = 3;'
```
dehume, line 18: If we can't test private link. Do we need to include this?

integration/connection.tf
```diff
@@ -16,4 +16,40 @@ resource "materialize_connection" "schema_registry" {
...
+  kafka_brokers = [
+    {
+      broker = "kafka:9092"
+    },
+    {
+      broker = "kafka2:9092"
+    }
+  ]
+  kafka_sasl_username   = "sasl_user"
+  kafka_sasl_password   = format("%s.%s.%s", materialize_database.database.name, materialize_schema.schema.name, materialize_secret.kafka_password.name)
```
dehume, line 42: Do you think we should add a fully qualified output for resources? Separate from this PR

.github/workflows/integration.yml
benesch: You can still do an integration test that _creates_ the PrivateLink connection (to make sure you got the connection syntax right)! It'll show up in `mz_connections` etc. You just won't be able to create a source that actually uses that PrivateLink connection.

bobbyiliev: I've tried that already but I end up with a TF panic:

```
| 2023-02-22T21:23:10.596Z [INFO]  provider.terraform-provider-materialize: 2023/02/22 21:23:10 [ERROR] could not create connection: CREATE CONNECTION materialize.public.privatelink_conn TO AWS PRIVATELINK (SERVICE NAME 'com.amazonaws.vpce.us-east-1.vpce-svc-0e123abc123198abc',AVAILABILITY ZONES ('use1-az2', 'use1-az6'));: timestamp=2023-02-22T21:23:10.596Z
```

Even though TF manages to create the privatelink connection, it still panics.

My guess is that we are returning an error and TF detects that and fails?

```
materialize=> CREATE CONNECTION materialize.public.privatelink_conn2 TO AWS PRIVATELINK (SERVICE NAME 'com.amazonaws.vpce.us-east-1.vpce-svc-0e123abc123198abc',AVAILABILITY ZONES ('use1-az2', 'use1-az6'));
ERROR:  AWS PrivateLink connections are not supported

materialize=> show connections;
            name             |           type            
-----------------------------+---------------------------
 privatelink_conn            | aws-privatelink
 privatelink_conn2           | aws-privatelink
 ```

benesch: Ohh, shoot, right, to test PrivateLink connections you need to run with the Kubernetes orchestrator (`--orchestrator=kubernetes`). Bummer. We could get that working with some elbow grease, but it's probably not worth the trouble now.

bobbyiliev: Just removed the `ALTER SYSTEM SET max_aws_privatelink_connections = 3`

We could open a new issue for coming up with a way to test privatelink connections. Maybe we just use the DevEx env for that one test?

integration/connection.tf
bobbyiliev: Good point!

dehume (approved):

.github/workflows/integration.yml
bobbyiliev: Created an issue so that we could track this: https://github.com/MaterializeInc/terraform-provider-materialize/issues/41

If you two are ok with this, we could go ahead and merge this one for the moment

benesch: Don't block on me!

Issue #36, opened 2023-02-22
Break Resource Connections


dehume:
Instead of having a single resource for connections, break them up by type.

PR #39, opened 2023-02-23
PrivateLink: Use nested schemas

### LINKED ISSUE #30: Use nested schemas
I noticed that we're using Terraform's map type to represent Kafka brokers:

https://github.com/MaterializeInc/terraform-provider-materialize/blob/e20b557841486c9ea2dfc6c845796669555db46f/materialize/resources/resource_connection.go#L169

but then reaching inside the map to pull out expected fields:

https://github.com/MaterializeInc/terraform-provider-materialize/blob/e20b557841486c9ea2dfc6c845796669555db46f/materialize/resources/resource_connection.go#L564-L588

This is going to fail with an ugly panic if a user typos a map field in their HCL.

A better approach is to use nested schemas inside of a list. See the `ordered_cache_behavior` field in the `aws_cloudfront_distribution` resource for an example: https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/cloudfront_distribution

```
resource "aws_cloudfront_distribution" "s3_distribution" {
  # Cache behavior with precedence 0
  ordered_cache_behavior {
    path_pattern     = "/content/immutable/*"
    allowed_methods  = ["GET", "HEAD", "OPTIONS"]
    cached_methods   = ["GET", "HEAD", "OPTIONS"]
    target_origin_id = local.s3_origin_id

    forwarded_values {
      query_string = false
      headers      = ["Origin"]

      cookies {
        forward = "none"
      }
    }

    min_ttl                = 0
    default_ttl            = 86400
    max_ttl                = 31536000
    compress               = true
    viewer_protocol_policy = "redirect-to-https"
  }

  # Cache behavior with precedence 1
  ordered_cache_behavior {
    path_pattern     = "/content/*"
    allowed_methods  = ["GET", "HEAD", "OPTIONS"]
    cached_methods   = ["GET", "HEAD"]
    target_origin_id = local.s3_origin_id

    forwarded_values {
      query_string = false

      cookies {
        forward = "none"
      }
    }

    min_ttl                = 0
    default_ttl            = 3600
    max_ttl                = 86400
    compress               = true
    viewer_protocol_policy = "redirect-to-https"
  }
}
```

This allows Terraform's type checking to kick in, and filter out invalid fields, etc. 

Here's a similar example from the Frontegg provider that shows how this is implemented: https://github.com/frontegg/terraform-provider-frontegg/blob/9976fb90bb60aaa28fed945ee67aefb761bd8e23/provider/resource_frontegg_workspace.go#L544-L566

bobbyiliev:
As per the discussion in #30 removing the `kafka_broker` in favor of `kafka_brokers`.

Closes #30

### DIFF
--- a/examples/main.tf
+++ b/examples/main.tf
@@ -79,16 +79,16 @@ resource "materialize_connection" "privatelink_conn" {
   aws_privatelink_service_name       = "com.amazonaws.us-east-1.materialize.example"
   aws_privatelink_availability_zones = ["use1-az2", "use1-az1"]
 }
-resource "materialize_connection" "example_kafka_connection" {
-  name            = "example_kafka_connection"
+resource "materialize_connection" "example_kafka_privatelink_conn" {
+  name            = "example_kafka_privatelink_conn"
   connection_type = "KAFKA"
-  kafka_brokers {
+  kafka_broker {
     broker                 = "b-1.hostname-1:9096"
     target_group_port      = "9001"
     availability_zone      = "use1-az1"
     privatelink_connection = "privatelink_conn"
   }
-  kafka_brokers {
+  kafka_broker {
     broker                 = "b-2.hostname-2:9096"
     target_group_port      = "9002"
     availability_zone      = "use1-az2"
@@ -112,10 +112,10 @@ resource "materialize_connection" "example_postgres_connection" {
 resource "materialize_connection" "example_kafka_connection" {
   name            = "example_kafka_connection"
   connection_type = "KAFKA"
-  kafka_brokers {
+  kafka_broker {
     broker = "b-1.hostname-1:9096"
   }
-  kafka_brokers {
+  kafka_broker {
     broker = "b-2.hostname-2:9096"
   }
   kafka_sasl_username   = "example"
--- a/examples/main.tf
+++ b/examples/main.tf
@@ -72,27 +72,30 @@ resource "materialize_connection" "example_ssh_connection" {
 # # Create a AWS Private Connection
 # Note: you need the max_aws_privatelink_connections increased for this to work:
 # show max_aws_privatelink_connections;
-# resource "materialize_connection" "example_kafka_connection" {
-#   name            = "example_kafka_connection"
-#   connection_type = "KAFKA"
-#   # kafka_broker    = "example.com:9092"
-#   kafka_brokers = [{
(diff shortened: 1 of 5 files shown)


examples/main.tf
```diff
@@ -72,27 +72,30 @@ resource "materialize_connection" "example_ssh_connection" {
...
+  name                               = "privatelink_conn"
+  schema_name                        = "public"
+  connection_type                    = "AWS PRIVATELINK"
+  aws_privatelink_service_name       = "com.amazonaws.us-east-1.materialize.example"
+  aws_privatelink_availability_zones = ["use1-az2", "use1-az1"]
+}
+resource "materialize_connection" "example_kafka_connection" {
+  name            = "example_kafka_connection"
+  connection_type = "KAFKA"
+  kafka_brokers {
```
benesch, line 85: Oh, check me on this, but I think Terraform convention is to use the singular. So you'd call this `kafka_broker`, and then you can specify it between 0 and _N_ times!

benesch: (Although we should set `MinItems` to 1 to ensure that you specify it at least once.)

bobbyiliev: Good points! Just updated this

materialize/resources/resource_connection.go
```diff
@@ -301,7 +318,6 @@ type ConnectionBuilder struct {
 	postgresSSLKey                        string
 	postgresSSLMode                       string
 	postgresAWSPrivateLink                string
-	kafkaBroker                           string
 	kafkaBrokers                          []map[string]interface{}
```
benesch, line 321: Recommend introducing a 
```go
type kafkaBroker struct {
    hostname string
    targetGroupPort int
    whateverElse here
}
```

struct for internal use. It'll be more resilient to typos! Then you only need to make sure that the field name in the schema matches the _one_ place where you construct the `kafkaBroker` struct; everything else gets to use e.g. `kafka.hostname` which will be checked by the Go compiler.

bobbyiliev: Yes! I was planning to do that with the work in #36 but just committed that change

benesch (commented):
Just skimmed, so leaving the official approval to Dennis, but the parts I looked at looked 👌🏽

dehume (approved):

PR #40, opened 2023-02-23
Resource Qualified Name

dehume:
Include qualified name for resources to make it easier to reference resources in Terraform.


bobbyiliev (approved):
That was quick! Looks great!

PR #43, opened 2023-02-27
Design doc

dehume:
Include a `DESIGN.md` to outline the "philosophy" of the provider and decisions made.

### DIFF
--- a/DESIGN.md
+++ b/DESIGN.md
@@ -43,4 +43,7 @@ The id for all provider resources is the corresponding id in [mz_catalog](https:
 When initially creating a resource via SQL, the id is not returned as part of the command. That is why after we create a resource the provider will query the mz_catalog using the name (and if applicable schema and database names) to lookup the id which will then be set with the `ReadContext`.
 
 ### Dividing Resources
-Complex Materailize resources are separated out into more specific provider resources. For example sources are divided across `materialize_source_kafka`, `materialize_source_loadgen`, `materialize_source_postgres`. Resources that have a large number of possibly contradictory parameters should be given their own resource. This offers more guidance by allowing more accurate required parameters and not confusing users with details for unnecessary fields.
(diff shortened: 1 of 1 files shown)


dehume:
@bobbyiliev Feel free to add anything you think we are missing, then we can open this up to other folks.

DESIGN.md
```diff
@@ -0,0 +1,46 @@
...
+}
+```
+
+### Ids
+The id for all provider resources is the corresponding id in [mz_catalog](https://materialize.com/docs/sql/system-catalog/mz_catalog/). For example the id for sources can be found in `mz_sources`. This is used both for creating new resources and [importing existing resources into state](https://developer.hashicorp.com/terraform/cli/import).
+
+When initially creating a resource via SQL, the id is not returned as part of the command. That is why after we create a resource the provider will query the mz_catalog using the name (and if applicable schema and database names) to lookup the id which will then be set with the `ReadContext`.
+
+### Dividing Resources
+Complex Materailize resources are separated out into more specific provider resources. For example sources are divided across `materialize_source_kafka`, `materialize_source_loadgen`, `materialize_source_postgres`. Resources that have a large number of possibly contradictory parameters should be given their own resource. This offers more guidance by allowing more accurate required parameters and not confusing users with details for unnecessary fields.
```
benesch, line 46:
```suggestion
Complex Materialize resources are separated out into more specific provider resources. For example sources are divided across `materialize_source_kafka`, `materialize_source_loadgen`, `materialize_source_postgres`. Resources that have a large number of possibly contradictory parameters should be given their own resource. This offers more guidance by allowing more accurate required parameters and not confusing users with details for unnecessary fields.
```

benesch, line 46: Ooh, this makes me think of another naming philosophy I'd like to propose: that the names always match exactly with Materialize! For example I think it should be `materialize_source_load_generator`, to exactly match the SQL statement and also the type in the catalog (`load-generator`).

benesch (approved):

bobbyiliev (approved):
LGTM!

PR #45, opened 2023-02-28
Sink Redesign

dehume:
Breaking up sink resources into their specific types. Currently there are only sinks for Kafka so more so just renaming the existing sink.


bobbyiliev (approved):
LGTM!

Issue #22, opened 2023-02-15
Integration Tests - Sources


dehume:
Complete integration tests for sources:
- Load generator
- Postgres
- Kafka/Redpanda

PR #47, opened 2023-03-01
Connections Redeisgn

### LINKED ISSUE #36: Break Resource Connections

bobbyiliev:
Breaking up connection resources into their specific types. 

Open to suggestions about the naming, eg. `materialize_resource_conflient_schema_registry` seems a bit lengthy.

Closes #36


examples/main.tf
```diff
@@ -60,28 +60,25 @@ resource "materialize_secret" "example_secret" {
...
 # # Create a AWS Private Connection
 # Note: you need the max_aws_privatelink_connections increased for this to work:
 # show max_aws_privatelink_connections;
-resource "materialize_connection" "privatelink_conn" {
-  name                               = "privatelink_conn"
-  schema_name                        = "public"
-  connection_type                    = "AWS PRIVATELINK"
-  aws_privatelink_service_name       = "com.amazonaws.us-east-1.materialize.example"
-  aws_privatelink_availability_zones = ["use1-az2", "use1-az1"]
+resource "materialize_connection_aws_privatelink" "privatelink_conn" {
```
dehume, line 74: For consistency maybe name it `"materialize_connection_aws_privatelink" "example_aws_privatelink_conn"`

integration/connection.tf
```diff
@@ -1,37 +1,25 @@
...
 }
 
-resource "materialize_connection" "example_ssh_connection" {
-  name            = "ssh_example_connection"
-  schema_name     = "public"
-  connection_type = "SSH TUNNEL"
-  ssh_host        = "ssh_host"
-  ssh_user        = "ssh_user"
-  ssh_port        = 22
+resource "materialize_connection_ssh_tunnel" "example_ssh_connection" {
```
dehume, line 13: Just to differentiate from the examples, maybe we should leave "example" out of the integration test resources.

integration/connection.tf
```diff
@@ -44,16 +32,15 @@ resource "materialize_connection" "kafka_conn_multiple_brokers" {
...
-  postgres_port     = 5432
-  postgres_user     = "materialize"
-  postgres_password = format("%s.%s.%s", materialize_database.database.name, materialize_schema.schema.name, materialize_secret.postgres_password.name)
-  postgres_database = "postgres"
+resource "materialize_connection_postgres" "example_postgres_connection" {
+  name     = "example_postgres_connection"
+  host     = "postgres"
+  port     = 5432
+  user     = "materialize"
+  password = format("%s.%s.%s", materialize_database.database.name, materialize_schema.schema.name, materialize_secret.postgres_password.name)
```
dehume, line 40: You should be able to get this with the `qualified_name` of the secret.

materialize/resources/resource_connection_aws_privatelink.go
```diff
@@ -0,0 +1,209 @@
...
+}
+
+func (b *ConnectionAwsPrivatelinkBuilder) PrivateLinkAvailabilityZones(privateLinkAvailabilityZones []string) *ConnectionAwsPrivatelinkBuilder {
+	b.privateLinkAvailabilityZones = privateLinkAvailabilityZones
+	return b
+}
+
+func (b *ConnectionAwsPrivatelinkBuilder) Create() string {
+	q := strings.Builder{}
+	q.WriteString(fmt.Sprintf(`CREATE CONNECTION %s.%s.%s TO AWS PRIVATELINK (`, b.databaseName, b.schemaName, b.connectionName))
```
dehume, line 110: Out of scope for this PR but maybe we should make `qualified_name` a func on the builder type. Would be cleaner to just have that than the database, schema, name string parsing in the Create, Drop and Rename SQL generators.

materialize/resources/resource_connection_aws_privatelink.go
```diff
@@ -0,0 +1,209 @@
...
+	databaseName := d.Get("database_name").(string)
+
+	builder := newConnectionAwsPrivatelinkBuilder(connectionName, schemaName, databaseName)
+
+	if v, ok := d.GetOk("service_name"); ok {
+		builder.PrivateLinkServiceName(v.(string))
+	}
+
+	if v, ok := d.GetOk("availability_zones"); ok {
+		azs := v.([]interface{})
```
dehume, line 155: Do you need to initialize a new slice and iterate over the existing one? Would this work?
```
	if v, ok := d.GetOk("availability_zones"); ok {
		builder. PrivateLinkAvailabilityZones(v.([]string))
	}
```

materialize/resources/resource_connection_aws_privatelink_test.go
```diff
@@ -0,0 +1,63 @@
...
+			mz_connections.type
+		FROM mz_connections
+		JOIN mz_schemas
+			ON mz_connections.schema_id = mz_schemas.id
+		JOIN mz_databases
+			ON mz_schemas.database_id = mz_databases.id
+		WHERE mz_connections.id = 'u1';`, b)
+}
+
+// here are common ^
```
dehume, line 54: ?

materialize/resources/resource_connection_postgres.go
```diff
@@ -0,0 +1,395 @@
...
+
+func newConnectionPostgresBuilder(connectionName, schemaName, databaseName string) *ConnectionPostgresBuilder {
+	return &ConnectionPostgresBuilder{
+		connectionName: connectionName,
+		schemaName:     schemaName,
+		databaseName:   databaseName,
+	}
+}
+
+func (b *ConnectionPostgresBuilder) ConnectionName(connectionName string) *ConnectionPostgresBuilder {
```
dehume, line 167: Same as above, can remove these.

materialize/resources/resource_connection_postgres.go
```diff
@@ -0,0 +1,395 @@
...
+}
+
+func (b *ConnectionPostgresBuilder) Create() string {
+	q := strings.Builder{}
+	q.WriteString(fmt.Sprintf(`CREATE CONNECTION %s.%s.%s TO POSTGRES (`, b.databaseName, b.schemaName, b.connectionName))
+
+	q.WriteString(fmt.Sprintf(`HOST '%s'`, b.postgresHost))
+	q.WriteString(fmt.Sprintf(`, PORT %d`, b.postgresPort))
+	q.WriteString(fmt.Sprintf(`, USER '%s'`, b.postgresUser))
+	q.WriteString(fmt.Sprintf(`, PASSWORD SECRET %s`, b.postgresPassword))
```
dehume, line 244: This may need a conditional, ore should be set as a required parameter. According to the docs, it isn't required but not sure if that is true.

https://materialize.com/docs/sql/create-connection/#postgres-options

materialize/resources/resource_connection_ssh_tunnel.go
```diff
@@ -0,0 +1,218 @@
...
+func connectionSshTunnelCreate(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	conn := meta.(*sqlx.DB)
+
+	connectionName := d.Get("name").(string)
+	schemaName := d.Get("schema_name").(string)
+	databaseName := d.Get("database_name").(string)
+
+	builder := newConnectionSshTunnelBuilder(connectionName, schemaName, databaseName)
+
+	if v, ok := d.GetOk("host"); ok {
```
dehume, line 160: Can remove all these conditionals.

dehume (changes requested):
Looks good. A couple nits on naming and required fields and some things can be removed.

materialize/resources/resource_connection_confluent_schema_registry.go
```diff
@@ -0,0 +1,310 @@
...
+		Description: "The URL of the Confluent Schema Registry.",
+		Type:        schema.TypeString,
+		Optional:    true,
+	},
+	"ssl_ca": {
+		Description: "The CA certificate for the Confluent Schema Registry.",
+		Type:        schema.TypeString,
+		Optional:    true,
+	},
+	"ssl_cert": {
```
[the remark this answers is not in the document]
bobbyiliev: This is only required for SSL client authentication, if we make it required here, it will always be required, even when using password auth

materialize/resources/resource_connection_aws_privatelink.go
bobbyiliev: Yes, this is exactly how I initially had it, but then realized that it is hitting a nasty panic. eg:

```
panic: interface conversion: interface {} is []interface {}, not []string
```

This is only caught during the tf apply, the go tests don't catch it.
Hence the iteration, any suggestions for improvements would be appreciated!

bobbyiliev: Actually just working on the integration tests for the sources, and noticed the same error occurs for the Postgres Tables as well, eg:

```
| panic: interface conversion: interface {} is map[string]interface {}, not map[string]string
| 
| goroutine 173 [running]:
| terraform-materialize/materialize/resources.sourcePostgresCreate({0x93d928, 0x400010f2c0}, 0x0?, {0x7fa000?, 0x400014b1d0?})
|       /usr/src/app/materialize/resources/resource_source_postgres.go:247 +0x39c
```

Will try to submit a patch for that with the tests from #22

dehume: Alright. Let's create a separate issue for that for all list types. Thanks

dehume (approved):

PR #51, opened 2023-03-07
Interface Tests

dehume:
Making changes so resources can have better testing. Now unit tests cover both the different versions of the SQL that can be generated and actual initialization of create and delete commands.

### DIFF
--- a/materialize/resources/resource_secret.go
+++ b/materialize/resources/resource_secret.go
@@ -2,7 +2,6 @@ package resources
 
 import (
 	"context"
-	"database/sql"
 	"fmt"
 	"log"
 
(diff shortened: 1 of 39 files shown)


materialize/resources/resource_secret.go
```diff
@@ -122,19 +120,30 @@ func readSecretParams(id string) string {
...
-}
-
 func secretRead(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
 	conn := meta.(*sqlx.DB)
 	i := d.Id()
 	q := readSecretParams(i)
 
-	readResource(conn, d, i, q, _secret{}, "secret")
+	var name, schema, database string
+	if err := conn.QueryRowx(q).Scan(&name, &schema, &database); err != nil {
```
dehume, line 129: This has been driving me nuts. Ultimately moved away from using `StructScan` since I was having a really hard time getting the aliased columns to match up with the exposed fields. It would probably be safer to use a struct here but may have to circle back on it.

bobbyiliev (approved):

PR #56, opened 2023-03-08
Qualified name

dehume:
Add a qualified name function to the builder rather than create the qualified name field from the read. This allows us to use the qualified name in the SQL construction for consistency. Also enclosing the components of the qualified name in quotations for safety.


bobbyiliev (approved):
Neat! That was fast! Looks good!

PR #58, opened 2023-03-09
Postgres Tables

dehume:
Consistent use of `tables` across sources (postgres and load generator sources) and fix tables field in load generator.


bobbyiliev (approved):
Looks good!

PR #63, opened 2023-03-09
Resource Index

dehume:
Add support for index as a resource and data source.


dehume:
@bobbyiliev I feel like I'm not quite reading the `col_expr` correctly. Based on documentation it seems like it should be a struct with two fields
```
col_expr… | The expressions to use as the key for the index.
-- | --
field | The name of the option you want to set.
val | The value for the option.
```
But all the examples make me think it would just be a list of strings?

dehume:
Little confused on how to correctly delete an index. Tried it again locally using a different name for the index and still getting an error on the drop.
```
materialize=> SHOW INDEXES ON example.example.load_gen_cluster;
     name      |        on        | cluster | key
---------------+------------------+---------+-----
 loadgen_index | load_gen_cluster | cluster | {}
(1 row)

materialize=> DROP INDEX loadgen_index RESTRICT;
ERROR:  unknown catalog item 'loadgen_index'
```

benesch:
You need to fully qualify the name of the index, since it's in the `example.example` schema, which does not appear to be your currently active schema.

dehume:
May have to think out the best way to handle the qualified name. Since it depends on the object the index exists on. It seems like we should have `schema_name` and `database_named` be `Computed` from that. Didn't want to just parse the schema and database name from the `obj_name` field in case they don't qualify it so moved the logic into the queries. Only problem I'm having is the joining of the `on_id` since it can be any number of possible mz catalogue tables (sources, tables...). Can work that logic into the query but would be a little messier.

Am I missing a better solution to this?

materialize/resources/resource_index.go
```diff
@@ -0,0 +1,295 @@
...
+	"obj_name": {
+		Description: "The name of the source, view, or materialized view on which you want to create an index..",
+		Type:        schema.TypeString,
+		Required:    true,
+		ForceNew:    true,
+	},
+	"cluster_name": {
+		Description: "The cluster to maintain this index. If not specified, defaults to the active cluster.",
+		Type:        schema.TypeString,
+		Required:    true,
```
morsapaes, line 53:
```suggestion
		Optional:    true,
```

morsapaes: > If not specified, defaults to the active cluster.

dehume:
@bobbyiliev still not entirely ready but would like a sanity check on this. Based on the documentation it seems you don't fully qualify the index but to drop the index you do need the qualified name and that is take from the object the index points at. Changed the object to use the `IdentifierSchemaStruct` so it will have schema and database properties so we could get the schema, name from that or we could get it from that really ugly multi join query that joins `mz_indexes` against sources, views and materialized views. Either way it seems like indexes are going to behave slightly different.

dehume:
Alright after a few iterations and leveraging some of the nicer organization of the repo, the indexes will behave different from other resources but will inherit the schema and database from the object it looks at (as the dbt adaptor does). It will still have computed `schema_name` and `database_name` fields.

Only thing left that is still kind of ugly is the `ReadIndexParams` query which has to join out to all all the potential mz catalogue tables for what the index may be looking at.

bobbyiliev (approved):
I think that this looks very good! Maybe worth adding a couple of examples in the examples dir and a couple more integration tests (eg. an index on a table and a view)

Issue #54, opened 2023-03-08
Resource Materialized View


dehume:
We should have a terraform resource and datasource for materialized views

https://materialize.com/docs/sql/create-materialized-view/

Even if this would probably be better handled with the dbt adaptor

PR #64, opened 2023-03-10
Add view resource and datasource

### LINKED ISSUE #55: Resource View
We should have a terraform resource and datasource for views

https://materialize.com/docs/sql/create-view/

Even if this would probably be better handled with the dbt adaptor

bobbyiliev:
Adding view resource and datasource and also re-generated the docs.

Closes #55


materialize/resources/resource_view.go
```diff
@@ -0,0 +1,285 @@
...
+		Computed:    true,
+	},
+	"temporary": {
+		Description:   "Mark the view as temporary.",
+		Type:          schema.TypeBool,
+		Optional:      true,
+		ForceNew:      true,
+		ConflictsWith: []string{"or_replace"},
+	},
+	"if_not_exists": {
```
dehume, line 46: I don't think we need a `if_not_exists` or `or_replace` fields (the same also probably is true for `temporary` though I did include it in my table PR yesterday). The user shouldn't need to set these since those configurations should be managed by the Terraform provider and state.

materialize/resources/resource_view.go
```diff
@@ -0,0 +1,285 @@
...
+		Description:   "If specified, replace the view if it already exists.",
+		Type:          schema.TypeBool,
+		Optional:      true,
+		ForceNew:      true,
+		ConflictsWith: []string{"if_not_exists", "temporary"},
+	},
+	"select_stmt": {
+		Description: "The SQL statement to create the view.",
+		Type:        schema.TypeString,
+		Required:    true,
```
dehume, line 63: Should include a `ForceNew`

dehume (changes requested):
Looks good but I think we should remove a few fields from the Terraform schema.

materialize/resources/resource_view.go
bobbyiliev: Just added this

materialize/resources/resource_view.go
bobbyiliev: good points! I've removed this for the view resource, I will remove that for the tables in the next PR for #54

dehume (approved):
Looks good

PR #65, opened 2023-03-10
Add materialized view resource

bobbyiliev:
Adding materialized view resource and datasource and also re-generated the docs.

Closes https://github.com/MaterializeInc/terraform-provider-materialize/issues/54

Also removing the `temporary` field for `tables` as discussed in #64


materialize/resources/resource_materialized_view.go
```diff
@@ -0,0 +1,241 @@
...
+		Default:     "materialize",
+		ForceNew:    true,
+	},
+	"qualified_name": {
+		Description: "The fully qualified name of the materialized view.",
+		Type:        schema.TypeString,
+		Computed:    true,
+	},
+	"in_cluster": {
+		Description: "The cluster to maintain this materialized view.",
```
dehume, line 40: Should we add some info about the defaults for this? Either in the form of a default value or just in the description that it will be in the default cluster if not selected.

materialize/resources/resource_materialized_view.go
```diff
@@ -0,0 +1,241 @@
...
+}
+
+func (b *MaterializedViewBuilder) SelectStmt(selectStmt string) *MaterializedViewBuilder {
+	b.selectStmt = selectStmt
+	return b
+}
+
+func (b *MaterializedViewBuilder) Create() string {
+	q := strings.Builder{}
+	q.WriteString(`CREATE`)
```
dehume, line 101: Nit: But could be done in fewer lines, something like:
```
	q.WriteString(fmt.Sprintf(`CREATE MATERIALIZED VIEW %s`, b.qualifiedName()))

	if b.inCluster != "" {
		q.WriteString(fmt.Sprintf(` IN CLUSTER %s`, b.inCluster))
	}

	q.WriteString(fmt.Sprintf(` AS %s;`, b.selectStmt))
	return q.String()
``

dehume (approved):
Looks good. Two minor suggestions.

materialize/resources/resource_materialized_view.go
bobbyiliev: neat! thank you for this suggestion!

materialize/resources/resource_materialized_view.go
bobbyiliev: Just updated the description, let me know what you think

Issue #25, opened 2023-02-15
Secrets/Strings Parameters


dehume:
Certain resources can reference secrets or use strings for parameter values. For example when creating a [postgres connection](https://materialize.com/docs/sql/create-connection/#postgres) you can specify `SSL CERTIFICATE` either way. The SQL this translates to will change depending on the parameter type, using a secret will prepend `SECRET`.

For our resource parameters than can be configured both ways, we should have the same interface if users use a string or a resource reference.
```
ssl_certificate = "string"
```
or
```
ssl_certificate = materialize_secret.my_secret.name
```
Should both work.

benesch:
Does Terraform support fields that can take different types of objects like that? The way I've usually seen this done in Terraform resources is to have two separate fields:

```hcl
ssl_certificate = "literal string"
ssl_certificate_secret = <secret ref>
```

where the provider then enforces that you set one or the other, but not both.

One additional wrinkle here: referring to a secret in Terraform land is hard! The way the e.g. Snowflake provider handles object references is to take the database, schema, and name as three separate parameters. See, for example, the [`snowflake_table`](https://registry.terraform.io/providers/Snowflake-Labs/snowflake/latest/docs/resources/table_grant) resource. The way that would look here is (brace yourself):

```hcl
ssl_certificate = "literal string"
# OR
ssl_certificate_secret_database_name = "db"
ssl_certificate_secret_schema_name = "schema"
ssl_certificate_secret_name = "sekret"
```

I can't see a way to do it with a single `ssl_certificate_secret` field name that doesn't have a bunch of pitfalls around identifier quoting.

PR #66, opened 2023-03-12
Escape all identifiers and values

### LINKED ISSUE #27: Escape all identifiers and values

bobbyiliev:
Closes #27

### DIFF
--- a/materialize/resources/resource_schema.go
+++ b/materialize/resources/resource_schema.go
@@ -75,9 +75,9 @@ func (b *SchemaBuilder) ReadId() string {
 		SELECT mz_schemas.id
 		FROM mz_schemas JOIN mz_databases
 			ON mz_schemas.database_id = mz_databases.id
-		WHERE mz_schemas.name = '%s'
-		AND mz_databases.name = '%s';
-	`, b.schemaName, b.databaseName)
+		WHERE mz_schemas.name = %s
+		AND mz_databases.name = %s;
+	`, QuoteString(b.schemaName), QuoteString(b.databaseName))
 }
 
 func readSchemaParams(id string) string {
@@ -87,7 +87,7 @@ func readSchemaParams(id string) string {
 			mz_databases.name
 		FROM mz_schemas JOIN mz_databases
 			ON mz_schemas.database_id = mz_databases.id
-		WHERE mz_schemas.id = '%s';`, id)
+		WHERE mz_schemas.id = %s;`, QuoteString(id))
 }
 
 func schemaRead(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
--- a/materialize/resources/resource_schema.go
+++ b/materialize/resources/resource_schema.go
@@ -110,7 +110,7 @@ func schemaRead(ctx context.Context, d *schema.ResourceData, meta interface{}) d
 		return diag.FromErr(err)
 	}
 
-	qn := QualifiedName(database_name, name)
+	qn := fmt.Sprintf("%s.%s", database_name, name)
 	if err := d.Set("qualified_name", qn); err != nil {
 		return diag.FromErr(err)
 	}
--- a/materialize/resources/resource_schema.go
+++ b/materialize/resources/resource_schema.go
@@ -76,7 +76,7 @@ func (b *SchemaBuilder) ReadId() string {
 		FROM mz_schemas JOIN mz_databases
 			ON mz_schemas.database_id = mz_databases.id
 		WHERE mz_schemas.name = '%s'
-		AND mz_databases.name = '%s';	
+		AND mz_databases.name = '%s';
 	`, b.schemaName, b.databaseName)
 }
 
--- a/materialize/resources/utils.go
+++ b/materialize/resources/utils.go
@@ -69,7 +69,7 @@ func QuoteIdentifier(input string) (output string) {
 func QualifiedName(fields ...string) string {
 	var o []string
 	for _, f := range fields {
-		c := fmt.Sprintf(`"%v"`, f)
+		c := fmt.Sprintf(`%v`, QuoteIdentifier(f))
 		o = append(o, c)
 	}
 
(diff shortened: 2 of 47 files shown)


dehume (approved):

bobbyiliev:
Adding @benesch as a reviewer in case you have any suggestions/objections before we merge this

materialize/resources/resource_schema.go
```diff
@@ -76,7 +76,7 @@ func (b *SchemaBuilder) ReadId() string {
 		FROM mz_schemas JOIN mz_databases
 			ON mz_schemas.database_id = mz_databases.id
 		WHERE mz_schemas.name = '%s'
-		AND mz_databases.name = '%s';	
+		AND mz_databases.name = '%s';
 	`, b.schemaName, b.databaseName)
```
benesch, line 80: Don't these need escaping?

materialize/resources/utils.go
```diff
@@ -62,14 +62,18 @@ func QuoteString(input string) (output string) {
 }
 
 func QuoteIdentifier(input string) (output string) {
-	output = `"` + strings.Replace(input, `"`, `""`, -1) + `"`
+	parts := strings.Split(input, ".")
+	for i, p := range parts {
+		parts[i] = `"` + strings.Replace(p, `"`, `""`, -1) + `"`
+	}
+	output = strings.Join(parts, ".")
```
benesch, line 69: This indicates that something is getting conflated. Identifiers must _not_ split on `.`s, as you can say e.g.

```
CREATE TABLE "a.b.c" ...
```

which is quite different from:

```
CREATE TABLE "a"."b"."c" ...
```

Do you know what required this? Probably needs to be solved in another way.

benesch: Ahh, I see, it's the issue I brought up [here](https://github.com/MaterializeInc/terraform-provider-materialize/issues/25#issuecomment-1431892531):

> One additional wrinkle here: referring to a secret in Terraform land is hard! The way the e.g. Snowflake provider handles object references is to take the database, schema, and name as three separate parameters. See, for example, the [`snowflake_table`](https://registry.terraform.io/providers/Snowflake-Labs/snowflake/latest/docs/resources/table_grant) resource. The way that would look here is (brace yourself):
> 
> ```terraform
> ssl_certificate = "literal string"
> # OR
> ssl_certificate_secret_database_name = "db"
> ssl_certificate_secret_schema_name = "schema"
> ssl_certificate_secret_name = "sekret"
> ```
> 
> I can't see a way to do it with a single `ssl_certificate_secret` field name that doesn't have a bunch of pitfalls around identifier quoting.

I think we need to sort this out, or we will forever have weird edge cases in what sort of names are allowed and what sort of quoting rules you have to follow. 

One way to make this more ergonomic is to use nested schemas, so you'd say:

```
resource "materialize_source_postgres" "example_source_postgres" {
  name                = "source_postgres"
  schema_name         = "schema"
  postgres_connection {
    # Optionally choose a different schema. (Defaults to the source's schema?)
    # schema_name = "somewhere_else"
    name = "pgconn"
  }
}
```

bobbyiliev: Hm yes good point, in this this seems to be non-trivial, as if the user passes database.schema.table, then if we were to just quote it with `CREATE TABLE "database.schema.table"` that will result in a table named `database.schema.table` rather than a table in the `"database"."schema"` schema.

Will need to rethink this as at the moment if someone has a dot in their identifier name, it will mess up the logic. Open to suggestions.

bobbyiliev: Snowflake have a similar approach:

https://github.com/Snowflake-Labs/terraform-provider-snowflake/blob/main/pkg/snowflake/escaping.go#L39-L52

But they have an extra `MustCompile` check inside that `for` loop

materialize/resources/resource_schema.go
bobbyiliev: Good catch! Just pushed that change.

materialize/resources/utils.go
bobbyiliev: I've just pushed that change for the `postgres_connection`, let me know what you think! 

Basically, I've abstracted the nested schema into a separate func so that it can be re-used among all identifiers.

Once we agree upon a concept, I will carry over the change to the rest of the identifiers.

bobbyiliev: All identifiers have now been updated and use the new syntax outlined in @benesch example above! Let me know what you think.

benesch (approved):
Holy cow, I can't believe how quickly you turned this around. Looks great!!

PR #67, opened 2023-03-13
Add secrets and strings options

### LINKED ISSUE #25: Secrets/Strings Parameters

bobbyiliev:
Closes #25

### DIFF
--- a/materialize/resources/resource_connection_postgres.go
+++ b/materialize/resources/resource_connection_postgres.go
@@ -55,29 +55,7 @@ var connectionPostgresSchema = map[string]*schema.Schema{
 		Optional:    true,
 		Default:     5432,
 	},
-	"user": {
-		Description: "The Postgres database username.",
-		Type:        schema.TypeList,
-		Required:    true,
-		MaxItems:    1,
-		MinItems:    1,
-		Elem: &schema.Resource{
-			Schema: map[string]*schema.Schema{
-				"text": {
-					Description:   "The Postgres database username.",
-					Type:          schema.TypeString,
-					Optional:      true,
-					ConflictsWith: []string{"user.0.secret"},
-				},
-				"secret": {
-					Description:   "The Postgres database username secret.",
-					Type:          schema.TypeString,
-					Optional:      true,
-					ConflictsWith: []string{"user.0.text"},
-				},
-			},
-		},
-	},
+	"user": secretStringSchema("user", "The Postgres database username.", true, false),
 	"password": {
 		Description: "The Postgres database password.",
 		Type:        schema.TypeString,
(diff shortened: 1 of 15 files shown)


materialize/resources/resource_connection_postgres.go
```diff
@@ -123,6 +130,7 @@ type ConnectionPostgresBuilder struct {
 	postgresHost           string
 	postgresPort           int
 	postgresUser           string
+	postgresUserSecret     string
```
dehume, line 133: Doesn't Postgres have some other option secret fields:
* SSL CERTIFICATE AUTHORITY

It would also probably be good to add some new tests for postgres

dehume (commented):
I still find this slightly confusing. It is tough that we have fields that can both take in text or a secret and some fields which require a secret. In postgres we have:
* `user`
* `user_secret`
* `password`

All of which are `TypeString`. I wonder if secrets should be their own type with a single field just to really make it clear. If they exist as their own configuration block it might look like this:
```
resource "materialize_connection_postgres" "postgres_connection" {
  name = "postgres_connection"
  host = "postgres"
  port = 5432
  user = "postgres"
  password = {
    secret_name = materialize_secret.secret.name
  }
  database = "postgres"
}

resource "materialize_connection_postgres" "postgres_connection" {
  name = "postgres_connection"
  host = "postgres"
  port = 5432
  user_secret = {
    secret_name = materialize_secret.username.name
  }
  password = {
    secret_name = materialize_secret.password.name
  }
  database = "postgres"
}
```
Or maybe the description for `password` just needs to be updated that it requires a secret. Trying to make this as consistent as possible with the documentation where we can call out the types.
Screenshot 2023-03-13 at 9 56 05 AM

bobbyiliev:
Just updated this so now we have the following syntax:

```
# Create a Postgres Connection
resource "materialize_connection_postgres" "example_postgres_connection" {
  name     = "example_postgres_connection"
  host     = "instance.foo000.us-west-1.rds.amazonaws.com"
  port     = 5432
  user {
    text = "example"
  }
  password = "example"
}
```

That way the user can select explicitly if they want the value to be a `secret` or `text` as we have it in the docs.

Open to naming suggestions!

dehume (approved):

PR #68, opened 2023-03-13
Add terraform release workflow

### LINKED ISSUE #4: Publish Provider to Terraform Registry
To make it easier for users to install. We should publish the provider to the Terraform Registry. Will need to work in the versioning and publishing into a Github Action

https://developer.hashicorp.com/terraform/registry/providers/publishing

bobbyiliev:
Adding a GitHub workflow to publish the provider to the Terraform Registry on every Tag.

Closes #4 

Tagging @benesch as we would need to add the required secrets to the repo


benesch:
Secrets should be in place!

dehume (approved):

PR #71, opened 2023-03-13
Kafka source: make `INCLUDE` metadata options consistent

morsapaes:
While reading through the source code, I noticed that `INCLUDE HEADERS` is handled as `TypeBool`, while all other metadata options were handled as `TypeString`. These are all similar, and we should lead with treating these options as booleans (since aliasing is optional). As a follow-up, we can add a separate option to handle aliases.

In general, I think we need to revisit how we build the SQL statements in the provider. Though I'm not versed in Go, using `fmt.Sprintf()` seems vulnerable to SQL injection; it sounds like we should be using something like [this](https://www.calhoun.io/what-is-sql-injection-and-how-do-i-avoid-it-in-go/)? CC @benesch

### DIFF
--- a/materialize/resources/resource_source_kafka_test.go
+++ b/materialize/resources/resource_source_kafka_test.go
@@ -22,18 +22,18 @@ func TestResourceSourceKafkaCreate(t *testing.T) {
 		"item_name":                  "item",
 		"kafka_connection":           "kafka_conn",
 		"topic":                      "topic",
-		"include_key":                "key",
+		"include_key":                true,
 		"include_headers":            true,
-		"include_partition":          "parition",
-		"include_offset":             "offset",
-		"include_timestamp":          "timestamp",
+		"include_partition":          true,
+		"include_offset":             true,
+		"include_timestamp":          true,
 		"format":                     "AVRO",
 		"key_format":                 "AVRO",
 		"envelope":                   "UPSERT",
 		"schema_registry_connection": "csr_conn",
 		"value_strategy":             "avro_key_fullname",
-		// "primary_key":                []interface{}{"key_1", "key_2", "key_3"},
-		// "start_offset":               []interface{}{1, 2, 3},
+		//"primary_key":                []interface{}{"key_1", "key_2", "key_3"},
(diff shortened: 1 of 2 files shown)


materialize/resources/resource_source_kafka_test.go
```diff
@@ -22,27 +22,27 @@ func TestResourceSourceKafkaCreate(t *testing.T) {
...
 		"key_format":                 "AVRO",
 		"envelope":                   "UPSERT",
 		"schema_registry_connection": "csr_conn",
 		"value_strategy":             "avro_key_fullname",
-		// "primary_key":                []interface{}{"key_1", "key_2", "key_3"},
-		// "start_offset":               []interface{}{1, 2, 3},
-		"start_timestamp": 100,
+		//"primary_key":                []interface{}{"key_1", "key_2", "key_3"},
+		//"start_offset":               []interface{}{1, 2, 3},
+		"start_timestamp": -1000,
```
morsapaes, line 37: I'm actually not sure what the behaviour is if you specify a positive integer here. 🤔  Maybe it'll start from the beginning of the topic, rather than the end? Using a negative integer sounds less confusing and is likely the most common option for users. This is just a big ol' nit and I haven't double-checked the behaviour otherwise.

dehume: 👍 We can also include more validation in the parameter to ensure users only do negative

morsapaes: Positive numbers are definitely allowed! Looking through our [test drive](https://github.com/MaterializeInc/materialize/blob/1481c8c4a5160845be2cd5e8e38c53e246e10a73/test/testdrive/kafka-time-offset.td#L71), it seems like this will literally match an existing Kafka timestamp (which...makes sense). Negative values are relative timestamps.

morsapaes:
Ah, escaping identifiers is already covered in #27 (thanks for the heads-up, @bobbyiliev).

dehume:
Closing in favor of https://github.com/MaterializeInc/terraform-provider-materialize/pull/100

PR #77, opened 2023-03-20
Update readme.md file

### LINKED ISSUE #44: Update README
The current README for the provider is still boilerplate from the scaffolding. Update to something more specific to actually using the provider.

bobbyiliev:
Adding some basic information in the README.md file before we publish it to the TF registry. We can work on adding a detailed getting started guide but I think that we should keep the README.md with only the essential information. Open to suggestions!

As mentioned in #44 we will need to update some of the information once the provider has been published like the link to the registry and etc.

Closes #44

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 # Terraform Provider Materialize
 
-⚠️⚠️ **WARNING!** ⚠️⚠️ The Terraform Provider for Materialize is under
-active development.
+> **Warning**
+> The Terraform Provider for Materialize is under active development.
 
 This repository contains a Terraform provider for the [Materialize platform](https://cloud.materialize.com/).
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -10,33 +10,115 @@ This repository contains a Terraform provider for the [Materialize platform](htt
...
+    name          = "kafka_password"
+    database_name = "materialize"
+    schema_name   = "public"
+  }
+  sasl_mechanisms = "SCRAM-SHA-256"
+  progress_topic  = "example"
+}
+```
+
+Then, run apply the changes:
```
dehume, line 65: I think you can remove this section since you mention using the `terraform apply` in the paragraph above.

README.md
```diff
@@ -10,33 +10,115 @@ This repository contains a Terraform provider for the [Materialize platform](htt
...
 
-```sh
-$ make testacc
+output name {
+  value       = data.materialize_connection.all
+}
 ```
 
-### Adding dependencies
+Then, check the Terraform plan:
```
dehume, line 84: Again. I would lean towards leaving this out as more assumed knowledge.

README.md
```diff
@@ -10,33 +10,115 @@ This repository contains a Terraform provider for the [Materialize platform](htt
...
 
-This provider uses [Go modules](https://github.com/golang/go/wiki/Modules). Please see the Go documentation for the most up to date information about using Go modules.
+```bash
+terraform plan
+```
+
+### Importing existing resources
+
+You can import existing resources into your Terraform state using the `terraform import` command. For example, to import an existing connection named `kafka_connection`, first add the resource definition to your Terraform project:
```
dehume, line 92: Worth mentioning explicitly that ids are on the ids from the mz catalogue and link that https://materialize.com/docs/sql/system-catalog/mz_catalog/

dehume (approved):
Looks good. Overall I think we can keep this pretty minimal. Will also want to update the installation guide when we publish our first release to the registry.

PR #85, opened 2023-03-22
Centralize Fields

dehume:
Centralizing some of the shared `schema.Schema` for consistency. 

Also tweaking the `IdentifierSchema` to not require both required and default since they will just be the opposite.


bobbyiliev (approved):
Looks great!

PR #86, opened 2023-03-22
Refactor format spec schema

### LINKED ISSUE #83: Source and sink formats and envelopes
I think we can make the handling of source/sink formats and envelopes more ergonomic and Terraform-y. Here's a quick sketch:

```hcl

resource "materialize_source_kafka" "src" {
    key_format {
        avro {
            schema_registry_connection {
                // ....
            }
            strategy = "latest"
        }
    }

    value_format {
        avro {
            schema_registry_connection {
                // ....
            }
            strategy = "latest"
        }
    }
}
```

```hcl
resource "materialize_source_kafka" "src" {
    format {
        csv {
             columns = 4
        }
    }
}
```

And you'd only be allowed to specify one nested `avro` or `csv` or `json` schema per format.

The only wrinkle here is formats that don't take arguments, like `json` or `text`. How do those fit into the above proposal?

If we find a nice way to do parameter-less formats, we should use it for envelopes, too. The envelopes are simple idents right now, but `ENVELOPE UPSERT` is likely to soon grow at least a `key` parameter that specifies what upsert key to use.

bobbyiliev:
Closes #83


bobbyiliev:
Currently this only covers _source_ format specs, I'm currently working on sinks formats as well. Let me know if you have any early feedback! Example:

```hcl
  format {
    avro {
      schema_registry_connection {
        name          = materialize_connection_confluent_schema_registry.schema_registry.name
        schema_name   = materialize_connection_confluent_schema_registry.schema_registry.schema_name
        database_name = materialize_connection_confluent_schema_registry.schema_registry.database_name
      }
    }
  }
 ```

**Comment by benesch** [comment]
This is neat! So JSON would be like this?

```hcl
format {
    json = true
}
```

Seems nice!

bobbyiliev:
Yep exactly, `json` and `text` would both use that syntax.

Just pushed the change for the sinks as well. Let me know what you think.

Regarding the envelopes, should we open a separate issue and implement the change once the upsert envelope has parameters?

benesch:
> Regarding the envelopes, should we open a separate issue and implement the change once the upsert envelope has parameters or shall we go ahead and try to implement this already?

How does backcompat work? That's the big open question I have! If v1 of the provider ships

```hcl
resource "materialize_source_kafka" {
    # ...
    envelope = "upsert"
}
```

can we later ship v2 with this?

```hcl
resource "materialize_source_kafka" {
    # ...
    envelope {
        upsert {
            key = "blah"
        }
    }
}
```

Or are we locked in to the `envelope = "upsert"` syntax?

benesch:
All the format stuff you've got in here looks 👌🏽 , Bobby! I only skimmed though so you might want Dennis to take a pass too.

bobbyiliev:
I think that we should be able to change the syntax down the road, what do you think @dehume?

dehume (approved):

bobbyiliev:
Just refactored the sink and source envelopes to use the new syntax, so that we can build upon this down the road.

Example:

```
resource "materialize_source_kafka" {
    # ...
    envelope {
        upsert = true
    }
}
```

PR #90, opened 2023-03-27
Qualified SQL Name

dehume:
Rename the resource field `qualified_name` to `qualified_sql_name`. Also for consistency, every builder resource has a `QualifiedName` function so every SQL queries should use that function.

And fixing the remaining `QuoteString` issues.


bobbyiliev (approved):
Looks good!

Issue #82, opened 2023-03-21
Small schema nits


benesch:
Didn't want to spam with dozens of issues, so figured I'd just collect a list of small items here:

  * [x] The `materialized_view` resource has a `select_stmt` resource but to match Snowflake's provider perhaps we should use `statement` instead?
  * [x] Ditto for `view`.
  * [x] Ditto for `sink`.
  * [x] The `secret` docs say that the value must be implicitly castable to bytea, but because we're forcing a string literal, it'd be more accurate to say that the value _must_ be a bytea string literal: https://materialize.com/docs/sql/types/bytea/#syntax.
  * [x] Do the connection resources need a `connection_type` read-only property? It's implied by the resource type itself. Vestigial from when there was only one connection resource type?
  * [x] Ditto for sources.
  * [x] The `primary_key` option should be called `unenforced_primary_key` to make clear that it's not actually checked by Materialize. Also isn't this an unsafe mode-only option?
  * [x] In the Kafka source, `include_headers` is a bool but everything else is a string indicating the alias.
  * [x] The `postgres_connection` field for PostgreSQL sources should probably just be called `connection` since the "PostgreSQL" is implied by the sourcetype.
  * [x] Ditto for Kafka.
  * [x] We're inconsistent about whether we use `cluster_name` (sources) or `in_cluster` (materialized views). I'd say `cluster_name` everywhere, for consistency with e.g. `database_name`.
  * [x] The PostgreSQL source `tables.name` field has the same identifier escaping problems we had elsewhere. Recommend breaking the name down into a nested resource with `schema_name`, `name`.
  * [x] Ditto for load generator sources.
  * [x] The load generator mixes together options from all load generator types. Recommend creating nested schemas for each type (`counter`, `auction`, `tpch`) and nesting all type-specific parameters accordingly.
  * [x] Go through all nested resources and make sure they all follow Terraform convention for field names (singular, not plural).

bobbyiliev:
My PR here covers everything except a couple of open points:

- _Do the connection resources need a `connection_type` read-only property? It's implied by the resource type itself. Vestigial from when there was only one connection resource type?_
    - Indeed we don't use this in the resources anymore, but we still use the `connection_type` read-only property in the connection datasource, eg: `data "materialize_connection" "all" {}`. 
-  _In the Kafka source, include_headers is a bool but everything else is a string indicating the alias._
[...]

PR #91, opened 2023-03-28
Small schema improvements

bobbyiliev:
Fixing a part of the some small schema nits as suggested in #82


dehume (commented):
Looks good for cluster name and statement. Maybe break off removing `connection_type` to a separate PR since there is a little more to untangle there besides just removing it from the builder in just the postgres resource.

dehume (approved):

PR #97, opened 2023-04-03
Loadgen Types

dehume:
Nesting specific load generator types in their own schema for better clarity. For example:
```
resource "materialize_source_load_generator" "load_generator_cluster" {
  ...
  load_generator_type = "COUNTER"
  tick_interval       = "500ms"
}
```
Would become:
```
resource "materialize_source_load_generator" "load_generator_cluster" {
  ...
  load_generator_type = "COUNTER"

  counter_options {
    tick_interval = "500ms"
  }
}
```

### DIFF
--- a/pkg/materialize/source_load_generator.go
+++ b/pkg/materialize/source_load_generator.go
@@ -87,43 +87,21 @@ func (b *SourceLoadgenBuilder) Create() string {
 	// Optional Parameters
 	var p []string
 
-	if b.loadGeneratorType == "COUNTER" {
-		if b.counterOptions.TickInterval != "" {
-			t := fmt.Sprintf(`TICK INTERVAL %s`, QuoteString(b.counterOptions.TickInterval))
-			p = append(p, t)
+	for _, t := range []string{b.counterOptions.TickInterval, b.auctionOptions.TickInterval, b.tpchOptions.TickInterval} {
+		if t != "" {
+			p = append(p, fmt.Sprintf(`TICK INTERVAL %s`, QuoteString(t)))
 		}
+	}
 
-		if b.counterOptions.ScaleFactor != 0 {
-			s := fmt.Sprintf(`SCALE FACTOR %.2f`, b.counterOptions.ScaleFactor)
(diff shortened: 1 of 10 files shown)


pkg/resources/resource_source_load_generator.go
```diff
@@ -119,28 +165,67 @@ func sourceLoadgenCreate(ctx context.Context, d *schema.ResourceData, meta any)
 		builder.LoadGeneratorType(v.(string))
 	}
 
-	if v, ok := d.GetOk("tick_interval"); ok {
-		builder.TickInterval(v.(string))
-	}
+	if v, ok := d.GetOk("counter_options"); ok {
```
dehume, line 168: This is ugly code. Trying to think of a more elegant way to coalesce the values while still being safe.

dehume (commented):
I'm getting a little thrown by the docs vs what seems to be allowed in the integration tests. Is this the correct way to think of the [options](https://materialize.com/docs/sql/create-source/load-generator/)?

| type | tick interval | scale factor | max cardinality | tables |
| --- | --- | --- | --- | --- |
| Counter | string | float | int | |
| Auction | string | float | | X |
| TPCH | string | float | | X |

pkg/materialize/source_load_generator.go
```diff
@@ -10,15 +10,31 @@ type TableLoadgen struct {
 	Alias string
 }
 
+type CounterOptions struct {
+	TickInterval   string
+	MaxCardinality bool
+}
```
bobbyiliev, line 16: Indeed this is nontrivial. I usually use the testdrive for some references but in this case it does not cover all of the cross options: 
https://github.com/MaterializeInc/materialize/blob/main/test/testdrive/load-generator.td

For example with the counter you can actually specify a scale factor: `Expected one of SCALE or TICK or MAX`, but in the docs we only explicitly mention the TPCH loadgen.

I think that the table that you've shared is spot on!

bobbyiliev (approved):
Looks great!

PR #98, opened 2023-04-03
Singular Field Names

dehume:
Keep TF field names singular

### DIFF
--- a/integration/connection.tf
+++ b/integration/connection.tf
@@ -34,8 +34,8 @@ resource "materialize_connection_kafka" "kafka_conn_multiple_brokers" {
     database_name = materialize_secret.kafka_password.database_name
     schema_name   = materialize_secret.kafka_password.schema_name
   }
-  sasl_mechanism = "SCRAM-SHA-256"
-  progress_topic = "progress_topic"
+  sasl_mechanisms = "SCRAM-SHA-256"
+  progress_topic  = "progress_topic"
 }
 
 resource "materialize_connection_postgres" "postgres_connection" {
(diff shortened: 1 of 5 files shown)


integration/connection.tf
```diff
@@ -34,8 +34,8 @@ resource "materialize_connection_kafka" "kafka_conn_multiple_brokers" {
     database_name = materialize_secret.kafka_password.database_name
     schema_name   = materialize_secret.kafka_password.schema_name
   }
-  sasl_mechanisms = "SCRAM-SHA-256"
-  progress_topic  = "progress_topic"
+  sasl_mechanism = "SCRAM-SHA-256"
+  progress_topic = "progress_topic"
```
bobbyiliev, line 38: I think that this is how we have it in the docs:

https://materialize.com/docs/sql/create-connection/#kafka-auth

It only accepts 1 parameter but is plural

pkg/resources/resource_source_kafka.go
```diff
@@ -45,7 +45,7 @@ var sourceKafkaSchema = map[string]*schema.Schema{
 		Optional:    true,
 		ForceNew:    true,
 	},
-	"include_headers": {
+	"include_header": {
```
bobbyiliev, line 48: Same as the sasl mechanisms, in the docs it is `INCLUDE HEADERS`:

https://materialize.com/docs/sql/create-source/kafka/#headers

dehume:
Reverted the two you mentioned, so now the only change is the nested `column` (which is closer to the original objective of the issue). All the other changes are just terraform formatting.

bobbyiliev (approved):
LGTM!

PR #99, opened 2023-04-03
Rename Connection Fields

dehume:
Rename `x_connection` fields to just be `connection` (as it is implied by the name resource type).

### DIFF
--- a/integration/source.tf
+++ b/integration/source.tf
@@ -21,7 +21,7 @@ resource "materialize_source_load_generator" "load_generator_cluster" {
 resource "materialize_source_postgres" "example_source_postgres" {
   name = "source_postgres"
   size = "2"
-  postgres_connection {
+  connection {
     name          = materialize_connection_postgres.postgres_connection.name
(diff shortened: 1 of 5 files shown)


dehume:
Cannot use `connection` as it is a [reserved word](https://github.com/hashicorp/terraform-plugin-sdk/blob/main/helper/schema/resource.go#L17).

benesch:
Oh bummer! Some options off the top of my head:

  1. Stick with the current approach.
  2. Add a noise word (`source_connection`, `external_connection`, `materialize_connection`).
  3. Add a trailing underscore (`connection_`).
  4. Use an abbreviation (`conn`).

bobbyiliev:
Ah yes, I hit the same problem when I tested this out as per my comment [here](https://github.com/MaterializeInc/terraform-provider-materialize/issues/82#issuecomment-1486480819).

I personally think that option 1 (Stick with the current approach) would the least confusing for the end users.

PR #116, opened 2023-04-20
Add principal property to aws privatelink conn

### LINKED ISSUE #115: `materialize_connection_aws_privatelink` include `principal` property
From discussion https://github.com/MaterializeInc/www/issues/366#issuecomment-1516220940. Include `principal` as a computed property on the `materialize_connection_aws_privatelink` resource.

```
SELECT principal
   FROM mz_aws_privatelink_connections plc
   JOIN mz_connections c ON plc.id = c.id
   WHERE c.name = 'privatelink_svc';
```

The easy solution would be to include this additional join as part of [ReadConnectionParams](https://github.com/MaterializeInc/terraform-provider-materialize/blob/main/pkg/materialize/connection.go#L37) since all connections share the same underlying connection queries. But we should probably break out materialize_connection_aws_privatelink to have its own queries.

NOTE: The integration test is its own hurdle and we can keep that to https://github.com/MaterializeInc/terraform-provider-materialize/issues/41

bobbyiliev:
Closes #115

### DIFF
--- a/docs/resources/source_load_generator.md
+++ b/docs/resources/source_load_generator.md
@@ -14,12 +14,15 @@ A source describes an external system you want Materialize to read data from, an
 
 ```terraform
 resource "materialize_source_load_generator" "example_source_load_generator" {
-  name                = "source_load_generator"
-  schema_name         = "schema"
-  size                = "3xsmall"
-  load_generator_type = "COUNTER"
-  tick_interval       = "500ms"
-  scale_factor        = 0.01
+  name        = "source_load_generator"
+  schema_name = "schema"
+  size        = "3xsmall"
+
+  counter_options {
+    load_generator_type = "COUNTER"
+    tick_interval       = "500ms"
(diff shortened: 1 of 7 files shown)


docs/resources/source_load_generator.md
```diff
@@ -14,12 +14,15 @@ A source describes an external system you want Materialize to read data from, an
...
-  size                = "3xsmall"
-  load_generator_type = "COUNTER"
-  tick_interval       = "500ms"
-  scale_factor        = 0.01
+  name        = "source_load_generator"
+  schema_name = "schema"
+  size        = "3xsmall"
+
+  counter_options {
+    load_generator_type = "COUNTER"
```
dehume, line 22: I actually screwed this up in my fix. `load_generator_type` should be outside the `counter_options`

dehume (approved):
Looks good

PR #118, opened 2023-04-20
Remove unnecessary type property

dehume:
Remove the unnecessary type property from source, sink and connection resources. Was not part of the read query.


bobbyiliev (approved):
Looks good! We should include this in the `v0.0.3` release!

PR #119, opened 2023-04-21
Cleanup `resources` Functions

dehume:
Doing some cleanup of the `resources`. There were cases where we had logic defined within the resources on how to set certain nested structs and it was defined in multiple places. Centralizing all that logic in the `materialize` package to try and keep the `resources` package specific to Terraform. Also a few other odds and ends cleanup.


bobbyiliev (approved):
Looks great!

PR #120, opened 2023-04-25
Update Github Action Schedules

dehume:
Changing Dependabot to weekly on Sunday since daily is unnecessary. Likewise changing the scheduled integration tests to once a week since there is no point running the cron for those tests unless there is a new `materialize/materialized` image.


bobbyiliev (approved):
LGTM!

PR #127, opened 2023-04-28
Include SSH keys in SSH connection

dehume:
Include the ssh keys as computed properties on the ssh connection resource. Currently keeping the two keys as separate `TypeString` properties. Not sure if it would make more sense to have that exist as a single property `TypeList`.


bobbyiliev (approved):
This is great! Thank you for adding it so quickly!

PR #131, opened 2023-04-28
Resource Role

dehume:
Even though `role` have limited use, including the bones for the resource. Added:
* materialize package builder
* data source
* resource

To Do:
* Add as part of integration tests


bobbyiliev (approved):
Looks great!

PR #137, opened 2023-05-04
Add mz egress ips datasource

### LINKED ISSUE #136: New Data Source: `mz_egress_ips`
In some cases you might want to get the Materialize Egress IPs ([`mz_egress_ips`](https://materialize.com/docs/sql/system-catalog/mz_catalog/#mz_egress_ips)) and allow them in your security group. To do that you have to `select * from mz_egress_ips;`.

It will be nice if we had this as a data source!

bobbyiliev:
Closes #136 

Example output:

```
ips = {
  "egress_ips" = tolist([
    "123.123.123.123",
    ...
  ])
  "id" = "egress_ips"
}
```


pkg/materialize/egress_ips.go
```diff
@@ -0,0 +1,15 @@
+package materialize
+
+type EgressIpsBuilder struct {
```
dehume, line 3: We don't actually need a Builder for this since we are just doing the data source query. So I think this just needs the `ReadEgressIpsDatasource` function which is a little silly in this case but probably good to just keep consistency.

dehume (approved):
One quick change and then this looks good.

dehume (approved):

PR #140, opened 2023-05-11
Resource Owner

dehume:
Adding resource for [ownership](https://materialize.com/docs/sql/alter-owner/). Ownership behaves a little different than the other resources (a lot of this will also apply to [grants](https://github.com/MaterializeInc/terraform-provider-materialize/issues/135)) so also doing some refactoring that might be good to apply to the other resources at some point in the future.

* Have the object builders execute the SQL rather than just returning strings. This will lead somewhat into the next point but this helped clean things up and prevent too much logic from leaking into Terraform layer and the `resources` package.
* This resource will require generating a custom id (will also apply to grants).
* This is only specific to `ownership` but the `DeleteContext` only removes the id from state, since there is no concept of dropping ownership.

### DIFF
--- a/pkg/materialize/constants.go
+++ b/pkg/materialize/constants.go
@@ -14,6 +14,7 @@ type ObjectType struct {
 	CatalogTable string
 }
 
+// https://materialize.com/docs/sql/grant-privilege/#details
 var ObjectPermissions = map[string]ObjectType{
 	"DATABASE": {
 		Permissions:  []string{"U", "C"},
--- a/pkg/materialize/constants.go
+++ b/pkg/materialize/constants.go
@@ -1,0 +1,68 @@
+package materialize
+
+var Permissions = map[string]string{
+	"r": "SELECT",
+	"a": "INSERT",
+	"w": "UPDATE",
+	"d": "DELETE",
+	"C": "CREATE",
+	"U": "USAGE",
+}
+
+type ObjectType struct {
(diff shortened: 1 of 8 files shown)


pkg/materialize/constants.go
```diff
@@ -0,0 +1,69 @@
...
+	"U": "USAGE",
+}
+
+type ObjectType struct {
+	Permissions  []string
+	CatalogTable string
+}
+
+// https://materialize.com/docs/sql/grant-privilege/#details
+var ObjectPermissions = map[string]ObjectType{
```
dehume, line 18: All the permissions will be used by grants

pkg/materialize/ownership.go
```diff
@@ -0,0 +1,118 @@
...
+			o.owner_id,
+			r.name
+		FROM %s o
+		JOIN mz_roles r
+			ON o.owner_id = r.id
+		WHERE o.id = %s
+	`, o.CatalogTable, QuoteString(catalogId))
+
+	var i, r string
+	if err := b.conn.QueryRowx(q).Scan(&i, &r); err != nil {
```
dehume, line 110: This would be cleaner with `StructScan` but can't get it to work properly, not sure if that is Materialize thing or not.

bobbyiliev (approved):
Neat! Based on a quick skim this looks great!

PR #141, opened 2023-05-12
Update Indexes Query

dehume:
Simplify the indexes queries to use `mz_objects` rather than rely on an overly complicated multi-join query.


bobbyiliev (approved):
LGTM!

PR #153, opened 2023-05-19
Param Struct Scan

dehume:
Use structs for all queries that lookup resource parameters. Generally safer and easier to read than just using variables. Applies to all resources. Also renaming files in the `resources` package to be prefaced with `resource_` for consistency.


bobbyiliev:
I can take a look at this as well!

bobbyiliev (approved):
Looks good!

PR #155, opened 2023-05-20
Refresh table state

bobbyiliev:
As reported by @chuck-alt-delete if we drop a table defined in Terraform  with psql and then reapply we get:

```
│ Error: sql: no rows in result set
│ 
│   with module.auction-house.materialize_table.winners_already_notified,
│   on auction-house/main.tf line 36, in resource "materialize_table" "winners_already_notified":
│   36: resource "materialize_table" "winners_already_notified" {
```

We could abstract this into a function and re-use it for other components as well.


pkg/resources/resource_table.go
```diff
@@ -68,7 +68,12 @@ func tableRead(ctx context.Context, d *schema.ResourceData, meta interface{}) di
 
 	var name, schema, database *string
 	if err := conn.QueryRowx(q).Scan(&name, &schema, &database); err != nil {
-		return diag.FromErr(err)
+		if err.Error() == "sql: no rows in result set" {
```
dehume, line 71:
```suggestion
		if err != sql.ErrNoRows {
```

dehume (approved):
Overall looks good. We can create a separate ticket to apply this to other resources

dehume:
Ticket for additional resources: https://github.com/MaterializeInc/terraform-provider-materialize/issues/157

PR #158, opened 2023-05-22
Base query

dehume:
This may be bordering on a little too quite. But now that we use structs for the queries, it is probably easiest just to maintain a single query for each query rather than separate ones for the id, parameter and datasource reads. Then if we move SQL execution into the `materialize` package, both datasources and resources can use the builder and the same underlying query but with different predicates.

POC using cluster replicas.


bobbyiliev (approved):
Looks good!

PR #159, opened 2023-05-22
Fix AZ Computed

dehume:
Make `availability_zone` computed for cluster replicas. The issue is if availability zones are not set in Terraform for replicas, one will be assigned at creation which will lead to a diff in state. Was not an issue with the local image since no availability zone is set.

Also turning on availability zones for the Materialize image so this will be caught in future integration tests.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -6,6 +6,9 @@ services:
   materialized:
     image: materialize/materialized:latest
     container_name: materialized
+    command:
+      - --availability-zone=test1
+      - --availability-zone=test2
     ports:
       - 6875:6875
       - 6878:6878
--- a/integration/cluster_replica.tf
+++ b/integration/cluster_replica.tf
@@ -5,10 +5,10 @@ resource "materialize_cluster_replica" "cluster_replica_1" {
 }
 
 resource "materialize_cluster_replica" "cluster_replica_2" {
-  name         = "r2"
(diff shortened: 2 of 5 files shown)


chuck-alt-delete (commented):
Thanks! I don't know enough about the terraform provider API to review this properly, and I don't have an easy way to test locally, but other than that, it looks good to me!

The thing I notice is `Compute: true` for the AZ now, which I think means terraform will assign the AZ using what it knows about the state of the system rather than forcing it to match what the user provides in the tf file. This is what we want! 

Thanks again!

benesch:
Does this work properly if manually specifying the AZ? How does that work
with computed properties? Particularly curious about the case where the
user specifies an AZ, plans, then *changes* the AZ.

dehume:
> Does this work properly if manually specifying the AZ? How does that work with computed properties? Particularly curious about the case where the user specifies an AZ, plans, then *changes* the AZ.

Yep. In the integration tests one replica is specified and one is not. The specified one uses `test2` and the not explicit one picks up `test1`. Then after the first apply, plan and apply show no changes. Updating the availability zone will still trigger a drop and replace.

bobbyiliev (approved):
> Yep. In the integration tests one replica is specified and one is not. The specified one uses test2 and the not explicit one picks up test1. Then after the first apply, plan and apply show no changes. Updating the availability zone will still trigger a drop and replace.

I think that this sounds good!

PR #162, opened 2023-05-24
Fix Kafka Conn Example

dehume:
Correct `sasl_username` in example for `materialize_connection_kafka`


bobbyiliev (approved):
Ah, good catch!

PR #166, opened 2023-05-26
Update `ValueSecretSchema` documentation

dehume:
Adding some more documentation to the `ValueSecretSchema` to make it slightly more clear. Also realized that the sink, source, connections resources still have generic descriptions from one when they were treated as single resources.


bobbyiliev (approved):
This looks great! Thank you for working on it!

Issue #165, opened 2023-05-25
TF Acceptance Tests


dehume:
Currently the integration tests with the docker compose using the Materialize image only does `apply` -> `plan` -> `destroy`. It would be good to also tests applicable updates and have a flow of: `apply` -> `plan` -> `update` -> `plan` -> `destroy`.

benesch:
Noticed this this weekend while I was working on #176. Terraform has a builtin acceptance test framework that I think we should use! I put up #177 for review.

dehume:
Remaining resources that will need acceptance tests after https://github.com/MaterializeInc/terraform-provider-materialize/pull/177 is merged.

Resource | Create | Delete | Update | Assigned | Notes |
| --- | --- | --- | --- | --- | --- |
| Cluster | X | X | | Done | Implemented with https://github.com/MaterializeInc/terraform-provider-materialize/pull/177
| Cluster Replica | X | X | X | @dehume | Implemented with https://github.com/MaterializeInc/terraform-provider-materialize/pull/177 (needs Update) |
| Connection AWS Privatelink |  X | X | X | @dehume |
| Connection Confluence Schema |  X | X | X | @dehume |
| Connection Kafka |  X | X | X | @dehume |
| Connection Postgres |  X | X | X | @dehume |
| Connection SSH Tunnel |  X | X | X | @dehume |
| Database |  X | X | | @dehume |
| Index | X | X | | @dehume |
| Materialized View | X | X | X | @bobbyiliev |
| Schema | X | X | | @bobbyiliev |
| Secret | X | X | X | Done | Implemented with https://github.com/MaterializeInc/terraform-provider-materialize/pull/177 |
| Sink Kafka | X | X | X | @bobbyiliev |
| Source Kafka | X | X | X | @bobbyiliev |
|Source Load Generator | X | X | X | @bobbyiliev |
| Source Postgres |  X | X | X | @bobbyiliev |
| Table | X | X | X | @bobbyiliev |
| Type | X | X | | @bobbyiliev |
| View | X | X | X | @bobbyiliev |

dehume:
Will be covered in two PRs:
* https://github.com/MaterializeInc/terraform-provider-materialize/pull/200
* https://github.com/MaterializeInc/terraform-provider-materialize/pull/201

Issue #157, opened 2023-05-22
Refresh State No Rows


dehume:
Applied to [tables](https://github.com/MaterializeInc/terraform-provider-materialize/pull/155) we should generalize this and apply to other resources.

PR #177, opened 2023-06-05
Add acceptance tests

benesch:
Not going to have time to work on this further, but I figured it would be a useful start!

----

Terraform has a standard acceptance testing framework [0] that allows for testing the complete lifecyle of provider resources: apply, plan, refresh, destroy, etc. In particular, this makes it possible to test things like a resource going missing due to a manual drop (#157).

This commit wires up the acceptance test framework, and then adds some basic acceptance tests for creating secrets, creating cluster replicas, and refreshing state for disappearing secrets.

Ultimately I think we should move all of today's integration tests to this acceptance test framework.

Work towards #165.

[0]: https://developer.hashicorp.com/terraform/plugin/testing/acceptance-tests


dehume (approved):

dehume:
Playing around with this a little more. Should these acceptance tests have a build flag or anything so we can run them separate from the unit quicker unit tests that don't involve Docker?

benesch:
I think that's rigged up automatically via the `TFACC` env var! Although I
wonder if we need to do the *reverse* in `make testacc` to avoid
running unit tests in the acceptance test job.

dehume:
Added coverage in the secrets acceptance tests for updates and uncovered an issue which I think explains https://github.com/MaterializeInc/terraform-provider-materialize/issues/194. Unfortunately this is probably in a few other resources.

Going to go ahead and merge this today and then @bobbyiliev and I can knock out the [remaining acceptance tests](https://github.com/MaterializeInc/terraform-provider-materialize/issues/165#issuecomment-1583157979) and fix any outstanding update issues and get that out in the release later this week.

bobbyiliev (approved):

PR #184, opened 2023-06-07
Application Name Connection String

dehume:
Include `application_name` session variable when executing SQL with the Terraform provider. Making it a provider attribute with a default so it can be overwritten to something else in the Pulumi provider (otherwise this can just be hardcoded).


benesch (commented):
Is there a way to hide this from the docs? Would love to discourage people from overriding this.

dehume:
> Is there a way to hide this from the docs? Would love to discourage people from overriding this.

I think we could edit the `index.md.tmpl` template and replace the `{{ .SchemaMarkdown | trimspace }}` reference with one we just hard code to only include the parameters we want to make public (so everything that isn't application name).

benesch (approved):
This is great, thank you!

morsapaes (approved):
Thank you for also considering the Pulumi provider, @dehume!

PR #189, opened 2023-06-08
Attributes Missing ForceNew

dehume:
Some resources had `UpdateContext` implemented but included attributes that were not marked `ForceNew` which would not trigger updates on changes.

https://github.com/MaterializeInc/terraform-provider-materialize/pull/188


benesch (approved):
LGTM! I'm sure there are a few more that have slipped through the cracks, but the comprehensive upgrade tests that are in the works should smoke those out.

PR #191, opened 2023-06-09
Resource `Grant`

dehume:
Adds a new resource type grants to handle privileges on database objects within materialize. This implements the easiest and more manageable way that @alex-hunt-materialize suggested in the overall [design](https://www.notion.so/materialize/Terraform-RBAC-c53be2942d1646dea3782b0d439a7972). This implements a grant resource at the level of a single role, object and privilege.

## Resource
```
resource "materialize_grant" "dennis_database_create" {
  role_name = "dennis"
  privilege = "CREATE"

  object {
    type = "DATABASE"
    name = "materialize"
  }
}
```
would grant the role `dennis` CREATE on the database `materialize` on apply and revoke on destroy.
```
resource "materialize_grant" "dennis_table_select" {
  role_name = "dennis"
  privilege = "SELECT"

  object {
    type = "TABLE"
    name = "my_table"
    schema_name = "my_schema"
    database_name = "materialize"
  }
}
```
would grant the role `dennis` SELECT on table `materialize.my_schema.my_table` on apply and revoke on destroy.

Because there is a 1:1 relationship there can be no updates so `UpdateContext` is not implemented

## Id
Because there is no id or catalog table for grants. The id is a concatenation of the object type, object id, role id and privilege.  So the database example above would be `GRANT|DATABASE|u1|u1|CREATE`.

## Implementation
I could not manage querying privileges for all the various objects with a simple query. So instead the `Scan` functions for all resources now includes privileges. However, privileges are not stored in state for the individual resources. Then in the `materialize` package the `ScanPrivileges` will use the function of the given object type to correctly pull in privilege information. Similar `PrivilegeId` will use the `Id` function of the given object type to get the id of the object and generate the compound id outlined above.

NOTE: Most of the code in this PR is adding privilege to all the existing resources as well as cleaning up and consolidating the mock scans for the unit tests since it was getting to hard to manage those in multiple places.


benesch:
Nice! Just catching up on @alex-hunt-materialize's feedback now. Will the fact that resources that name the same object/role/privilege have the same ID cause Terraform to automatically detect and reject duplicate resources? E.g.

```
resource "materialize_grant" "dennis_table_select" {
  role_name = "dennis"
  privilege = "SELECT"

  object {
    type = "TABLE"
    name = "my_table"
    schema_name = "my_schema"
    database_name = "materialize"
  }
}

resource "materialize_grant" "dennis_table_select_oops_duplicated" {
  role_name = "dennis"
  privilege = "SELECT"

  object {
    type = "TABLE"
    name = "my_table"
    schema_name = "my_schema"
    database_name = "materialize"
  }
}
```

Also: did we discuss the merits of a single `materialize_grant` resource vs multiple grant resources for different object types (`materialize_database_grant`, `materialize_schema_grant`, `materialize_cluster_grant`, `materialize_table_grant`)? The Postgres provider uses the former while the Snowflake provider uses the latter. The biggest benefit of the latter seems to be that we could do a lot more validation at the Terraform provider layer—e.g., statically rejecting a `schema_name` on a `materialize_database_grant` resource, for example.

dehume:
>Will the fact that resources that name the same object/role/privilege have the same ID cause Terraform to automatically detect and reject duplicate resources? E.g.

I don't think it would handle duplicates gracefully but have to test that out.

> Also: did we discuss the merits of a single `materialize_grant` resource vs multiple grant resources for different object types (`materialize_database_grant`, `materialize_schema_grant`, `materialize_cluster_grant`, `materialize_table_grant`)? 

I hadn't really considered multiple grant resources. I'm more used to the single grant object but we can ask the group. While it isn't handled currently we should be able to add some checks in the `ValidateFunc`. I was playing around with that some in the `Ownership` resource but need better test coverage around it.

dehume:
Playing around with this, the validation may not be possible because it relies on other attributes and I am only seeing the ability to write validation functions for single attribute.

dehume:
Converted to what this might look like for specific object resources. Makes validation much easier because we can limit the schema for the resource object and can use `ValidateFunc` to ensure that only allowed privileges are set.

If this looks good, can finish applying this to all the other resources grants. Though curious if this also means that the `ownership` resource should be be broken up into an ownership resource for each object type.

benesch:
The separate resource types look good to me.

> Though curious if this also means that the `ownership` resource should be be broken up into an ownership resource for each object type.

Should ... `owner` just be a property on each resource type?

dehume:
> Should ... `owner` just be a property on each resource type?

That probably makes the most sense from the user's perspective. I can play around with that to see what it would look like. https://github.com/MaterializeInc/terraform-provider-materialize/issues/197

benesch:
> > Will the fact that resources that name the same object/role/privilege have the same ID cause Terraform to automatically detect and reject duplicate resources? E.g.
> 
> I don't think it would handle duplicates gracefully but have to test that out.

Did you end up getting an answer to this question?

pkg/resources/resource_grant_database.go
```diff
@@ -0,0 +1,139 @@
...
+		ForceNew:    true,
+	},
+	"privilege": {
+		Description:  "The privilege to grant to the object.",
+		Type:         schema.TypeString,
+		Required:     true,
+		ForceNew:     true,
+		ValidateFunc: validPrivileges("DATABASE"),
+	},
+	"object": {
```
benesch, line 27: Mentioned this on the Notion doc but figured it might get lost there. I feel like either we should follow Snowflake and make these top-level parameters...

```hcl
resource "materialize_grant_database" "..." {
    role_name = "..."
    privilege = "..."
    database_name = "..."
}
```

Or we should name the nested subresource after the object type rather than using `object`:

```hcl
resource "materialize_grant_database" "..." {
    role_name = "..."
    privilege = "..."
    database {
        name = "..."
    }
}
```

dehume:
Moved the attributes to the top level.

Also played around with resources that generate the same id. It looks like resources with the same id can exist in Terraform state. And selectively making changes to one will update the correct one so there must be some additional information Terraform uses to dedupe. So it would be valid to create multiple grant resources that have the same role, object and privilege though I think you would still get into trouble if you create two, deleted one which would issue a revoke and then lead to conflict, though on the next apply it would then try and reapply the grant but that seems to be getting into user error.  But it is something we can play around with more next week when we start to enable these things and get setup the acceptance tests and different permutations.

benesch:
> Also played around with resources that generate the same id. It looks like resources with the same id can exist in Terraform state. And selectively making changes to one will update the correct one so there must be some additional information Terraform uses to dedupe. So it would be valid to create multiple grant resources that have the same role, object and privilege though I think you would still get into trouble if you create two, deleted one which would issue a revoke and then lead to conflict, though on the next apply it would then try and reapply the grant but that seems to be getting into user error. But it is something we can play around with more next week when we start to enable these things and get setup the acceptance tests and different permutations.

In that case it seems like there is less of a benefit to to this one-privilege-per-Terraform-resource model. You still have to be careful not to grant the same privilege on the same object to the same role twice. I imagine that we'll get requests to have a resource that allows granting multiple privileges at the same time (still only to one role). E.g.:

```hcl
resource "materialize_grant_table" "..." {
    role_name = "..."
    privileges = ["SELECT", "INSERT"]
    table_name = "..."
    schema_name = "..."
    database_name = "..."
}
```

But we can cross that bridge when we get there. It's an easy extension to the current resource to support both `privilege` and `privileges`:

```hcl
resource "materialize_grant_table" "..." {
    role_name = "..."
    table_name = "..."
    schema_name = "..."
    database_name = "..."
    # Only one of the following properties may be set.
    privilege = "UPDATE"
    privileges = ["SELECT", "INSERT"]
}
```

dehume:
The main thing that will add complexity by allowing multiple privileges is that the resource will need to support updates. But I'm fine leaving that for later. Otherwise I'm fine with this for V1.

bobbyiliev (approved):

dehume:
Merging with followup PRs for:
* Additional grant resources
* Enabling those resources (where I'm sure there will be more fine tuning)

PR #193, opened 2023-06-12
Change Kafka source json format to bytes

bobbyiliev:
Changing the Kafka source JSON format to BYTES as discussed [here](https://materializeinc.slack.com/archives/C015RHB3LDR/p1686554388443089).

Leaving out format JSON for the moment as support for the more ergonomic FORMAT JSON is in progress ([#7186](https://github.com/MaterializeInc/materialize/issues/7186))!

### DIFF
--- a/integration/source.tf
+++ b/integration/source.tf
@@ -58,6 +58,20 @@ resource "materialize_source_kafka" "example_source_kafka_format_text" {
   }
 }
 
+resource "materialize_source_kafka" "example_source_kafka_format_bytes" {
+  name = "source_kafka_bytes"
+  size = "2"
+  kafka_connection {
+    name          = materialize_connection_kafka.kafka_connection.name
+    schema_name   = materialize_connection_kafka.kafka_connection.schema_name
+    database_name = materialize_connection_kafka.kafka_connection.database_name
+  }
+  topic = "topic1"
+  format {
+    bytes = true
+  }
+}
+
 resource "materialize_source_kafka" "example_source_kafka_format_avro" {
   name = "source_kafka_avro"
   size = "2"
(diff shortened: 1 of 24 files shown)


pkg/materialize/format_specs.go
```diff
@@ -21,7 +21,7 @@ type FormatSpecStruct struct {
 	Avro     *AvroFormatSpec
 	Protobuf *ProtobufFormatSpec
 	Csv      *CsvFormatSpec
-	Json     bool
+	Bytes    bool
```
morsapaes, line 24: Generic question unrelated to the PR, but is there a reason for us to not name this type `SourceFormatSpecStruct`, if we then use `SinkFormatSpecStruct` for sinks? @dehume

morsapaes (approved):

integration/source.tf
```diff
@@ -58,6 +58,20 @@ resource "materialize_source_kafka" "example_source_kafka_format_text" {
   }
 }
 
+resource "materialize_source_kafka" "example_source_kafka_format_bytes" {
+  name = "source_kafka_bytes"
+  size = "2"
```
sthm, line 63: This looks a bit funny, but it's also used for the avro source below. Is this actually working, or should this be `2xsmall`?

sthm (approved):

pkg/materialize/format_specs.go
dehume: For consistency, we probably should change it.

dehume (approved):

pkg/materialize/format_specs.go
bobbyiliev: Good point! Just updated this accordingly!

PR #202, opened 2023-06-14
Fix Index Object Schema

dehume:
The index acceptance test was flaky because there was an issue with the read. The object schema was mistakenly set to the object name. Fixed and updated `IndexParams` to have more descriptive fields.


bobbyiliev (approved):
LGTM!

PR #203, opened 2023-06-15
Update changelog for v0.0.8

dehume:
Adding changelog for `0.0.8` and also realized that in all the other dates we were using 2022.


bobbyiliev (approved):
Looks good!

PR #204, opened 2023-06-15
Allow RBAC for Integration Tests

dehume:
Tweak the Docker image to have RBAC enabled in preparation for enabling `roles` and `grant` resources.

NOTE: Needed to explicitly add the cluster for resources in the acceptance and integration tests. It may make sense to also add cluster as an optional attribute when connecting to the provider.


bobbyiliev (approved):
Looks good!

PR #208, opened 2023-06-19
Ownership Attribute

dehume:
Removing ownership as it's own resource type and instead adding it as an attribute on existing resources (using tables as an example).

This adds two new columns to the resource `ownership_role` and `ownership_id`. The ownership role is an optional field that the user can set. If set during creation, an `ALTER... OWNER` will be run immediately after the `CREATE`. Then as part of the read, the ownership id will be read in as a computed field. Even if the ownership role is not set, the ownership id is set as all objects have a default owner.

This will behave very similarly on updates. If the user changes the ownership role, an `ALTER... OWNER` is run and the new owner id is updated on the read. However since there is no revoking of ownership if the user removes the ownership role field, no alter is run and the state of the ownership role is set to nothing, though the owner id will still reflect the ownership maintained by Materialize.

Since we are now running a CREATE and ALTER as part of the create, there should probably be more sophisticated error handling. If you tried to say create a table with an ownership role using a role that does not exist, the create would be successful but the alter would not and how would we want to handle that failure.

NOTE: Changing the builder factories to take in an object rather than the string parameters of name, schema, database so that object can be more easily shared across the multiple builders used in the create resource.


pkg/resources/resource_table_test.go
```diff
@@ -40,30 +44,6 @@ func TestResourceTableCreate(t *testing.T) {
 			t.Fatal(err)
 		}
 	})
-
-}
-
-func TestResourceTableUpdate(t *testing.T) {
```
dehume, line 46: Removing the update tests since this is just much easier and more accurate with the acceptance tests.

pkg/resources/resource_table.go
```diff
@@ -111,30 +118,60 @@ func tableCreate(ctx context.Context, d *schema.ResourceData, meta interface{})
...
-		b := materialize.NewTableBuilder(meta.(*sqlx.DB), oldName.(string), schemaName, databaseName)
+		o := materialize.ObjectSchemaStruct{Name: oldName.(string), SchemaName: schemaName, DatabaseName: databaseName}
+		b := materialize.NewTableBuilder(meta.(*sqlx.DB), o)
 
 		if err := b.Rename(newName.(string)); err != nil {
 			return diag.FromErr(err)
 		}
 	}
 
+	if d.HasChange("owernship_role") {
```
benesch, line 157: Typo of "owernship." Here and in a few other places, too!

benesch:
Very nice!

> Then as part of the read, the ownership id will be read in as a computed field. Even if the ownership role is not set, the ownership id is set as all objects have a default owner.

This surprised me. The Terraform provider deals in names rather than identifiers approximately everywhere. Does something go wrong if we just join the `owner_id` against the roles table to pull out the role name? Then `ownership_role` is whatever you've set it to, unless you haven't set it to anything, in which case it's whatever role Materialize has automatically assigned.

dehume:
> Does something go wrong if we just join the owner_id against the roles table to pull out the role name?

I think my hangup was not wanting to change the base queries but adding another join is pretty easy. Removed the `ownership_id` field so there is just done with the role name.

dehume:
Incorporated the changes necessary from https://github.com/MaterializeInc/terraform-provider-materialize/pull/209 which was a big PR and also added ownership for clusters to see if the patterned would hold for another resource. I think this is good to merge and will move adding ownership for the other resources to a separate PR.

PR #209, opened 2023-06-20
Enable Grants

dehume:
Enable the grant resources in the provider and include grants in the integration and acceptance tests.

NOTE: Also correcting my misspelling of "privilege" which was at least consistently misspelled.


pkg/provider/acceptance_grant_table_test.go
```diff
@@ -0,0 +1,137 @@
...
+	})
+}
+
+func testAccGrantTableResource(roleName, tableName, schemaName, databaseName, privilege string) string {
+	return fmt.Sprintf(`
+resource "materialize_role" "test" {
+	name = "%s"
+}
+
+resource "materialize_database" "test" {
```
dehume, line 70: Not sure about including database and schema resources for these tests when we are only really after the table grants. Mostly doing it because the scan id functions require all the qualifiers and would otherwise require hardcoding.

dehume (commented):
Starting with grants for database, schema and tables to show what this looks like. It will be roughly the same for enabling all the other grants.

dehume:
Include remaining grants

bobbyiliev (approved):

PR #211, opened 2023-06-22
Enable Remaining Ownership Resources

dehume:


dehume:
Need to include the new acceptance tests for the new resources though everything follows a similar pattern so I'm not too worried.

bobbyiliev (approved):

PR #215, opened 2023-06-28
Index Columns Scan

dehume:
Include column scan as part of the `tableRead`


dehume:
Integration tests failing on the column type naming issue. Have to work in better aliases

bobbyiliev (approved):

PR #216, opened 2023-06-28
Add managed clusters support

bobbyiliev:
Preparing the initial draft PR for adding support for managed clusters.

To test this out at the moment we need to run the following:

```sql
ALTER SYSTEM SET enable_managed_clusters = true
```


morsapaes:
Managed clusters are now live in production, so this change can be merged when ready.

pkg/provider/provider.go
```diff
@@ -159,5 +160,17 @@ func providerConfigure(ctx context.Context, d *schema.ResourceData) (interface{}
 		return nil, diags
 	}
 
+	if testing {
+		_, err := db.Exec("ALTER SYSTEM SET enable_managed_clusters = true;")
```
dehume, line 164: If this is live in production, can we check if this can get removed from the testing connection?

morsapaes: Managed clusters are enabled by default for all environments. The flag was removed in [Materialize #20339](https://github.com/MaterializeInc/materialize/pull/20399).

bobbyiliev: Seems like the PR above is not yet merged so we still need that alter statement

pkg/resources/resource_cluster.go
```diff
@@ -15,6 +15,12 @@ import (
 var clusterSchema = map[string]*schema.Schema{
 	"name":           NameSchema("cluster", true, true),
 	"ownership_role": OwnershipRole(),
+	"replication_factor": {
```
dehume, line 18: Should `replication_factor` and `size` both include `RequiredWith` since you would need to set them together.

pkg/resources/schema.go
```diff
@@ -71,6 +71,15 @@ func SizeSchema(resource string) *schema.Schema {
 	}
 }
 
+func ManagedClusterSizeSchema(resource string) *schema.Schema {
```
dehume, line 74: Nit. Will this attribute be shared across multiple resources in the future or is it specific to clusters? If it is specific I would favor keeping it specific to the cluster schema.

pkg/resources/resource_cluster.go
```diff
@@ -63,6 +77,16 @@ func clusterCreate(ctx context.Context, d *schema.ResourceData, meta interface{}
 	o := materialize.ObjectSchemaStruct{Name: clusterName}
 	b := materialize.NewClusterBuilder(meta.(*sqlx.DB), o)
 
+	// replication factor
+	if v, ok := d.GetOk("replication_factor"); ok {
+		b.ReplicationFactor(v.(int))
```
dehume, line 82: Nit. We should probably only set replication factor if size is also valid.

dehume (approved):
A few small things but overall I think this makes sense.

pkg/resources/schema.go
bobbyiliev: I like that idea, just updated it accordingly!

pkg/resources/resource_cluster.go
bobbyiliev: good catch! just refactored this accordingly

PR #218, opened 2023-06-29
Update RBAC Resources

dehume:
Support for the updated RBAC functionality. Adds three new resources:

* `materialize_grant_system_privilege`
* `materialize_{object type}_grant_default_privilege`
* `materialize_role_grant`
* Renaming the existing grants to be `materialize_{object type}_grant`

Trying to get in sync with everything outline in the [rbac tutorial](https://materialize.com/docs/manage/access-control/rbac-tutorial/). That RBAC tutorial is now an integration test.


dehume:
Overview of everything is in the [rbac.md](https://github.com/MaterializeInc/terraform-provider-materialize/blob/983b837b5137904528115561b570e312b2c83842/rbac.md) doc.

benesch:
Sorry, extremely strapped for time this week and next, but here's a list of questions from a quick skim:

  * I just realized that we call our per-object grants e.g. `materialize_grant_table` but Snowflake calls them `materialize_table_grant`. This is already released, but given that RBAC hasn't been available possibly not too late to change? Or was it an intentional deviation?
  * The system privilege resource is called `materialize_grant_system` but the default privilege resource is called `materialize_grant_default_privilege`. Probably the former should be called `materialize_grant_system_privilege`?
  * The default privilege resource is following a different philosophy than the existing resources, which are per-object type. As annoying as this would be to type in, it seems like we might want to have a different default privilege resource for each object type? I've only thought about this for a few minutes so far, though.

dehume:
> * I just realized that we call our per-object grants e.g. `materialize_grant_table` but Snowflake calls them `materialize_table_grant`. This is already released, but given that RBAC hasn't been available possibly not too late to change? Or was it an intentional deviation?

Will change and include in the changelog

> * The system privilege resource is called `materialize_grant_system` but the default privilege resource is called `materialize_grant_default_privilege`. Probably the former should be called `materialize_grant_system_privilege`?

Will change

> * The default privilege resource is following a different philosophy than the existing resources, which are per-object type. As annoying as this would be to type in, it seems like we might want to have a different default privilege resource for each object type? I've only thought about this for a few minutes so far, though.

Can divide that out and see how it looks. Does make sense for consistency.

bobbyiliev (approved):
This looks good. Approving this as discussed!

PR #221, opened 2023-06-29
[WIP] Rollback Ownership Failures

dehume:
If a resource creation fails on the ownership transfer, rollback the resource. It would be good to DRY some of the resource operations up.


bobbyiliev (approved):
Looks good!

PR #223, opened 2023-07-03
Remove Role Attribute

dehume:
Remove role attributes per https://github.com/MaterializeInc/materialize/pull/20233. Will be set instead by system privilege resources.


dehume:
Temporarily disabling the passwords in the integration tests since the connection now waits for confirmation before being created.

PR #227, opened 2023-07-07
Support FORMAT JSON for sources

benesch:
Part of MaterializeInc/materialize#20387.


dehume (approved):

morsapaes:
Nit: it'd be helpful to adopt the best practice of adding an entry to `CHANGELOG.md` under `Unreleased` for any PR introducing new functionality (much like we do for the dbt adapter). @dehume @bobbyiliev It makes it easier to get an overview of what's coming in the next release (and saves you some time, then!).

PR #236, opened 2023-07-13
Validate Connection

dehume:
Include option for `WITH (VALIDATE = false)` for connections. This will allow us to turn back on certain integration tests.

https://github.com/MaterializeInc/materialize/pull/19487


bobbyiliev (approved):
Very nice!

Issue #233, opened 2023-07-13
Unable to grant role to user when the username contains a hyphen


aidan-melen:
**Provider Version**

`0.1.0`

**Terraform Version**

`1.5.2`

**Describe the bug**

Unable to grant role privileges to a user using the [materialize_role_grant](https://registry.terraform.io/providers/MaterializeInc/materialize/latest/docs/resources/role_grant) resource. Specifically when the `member_name` contains a hyphen.

```bash
╷
│ Error: ERROR: Expected end of statement, found operator "-" (SQLSTATE 42601)
│ 
│   with materialize_role_grant.role_grant_user,
│   on grant.tf line 5, in resource "materialize_role_grant" "role_grant_user":
│    5: resource "materialize_role_grant" "role_grant_user" {
│ 
╵
```

**Expected behavior**

This SQL works:

```sql
GRANT readonly TO "firstname-lastname@company.com";
```

so I would expect this terraform to also work:

```terraform
resource "materialize_role_grant" "role_grant_user" {
  role_name   = "readonly"
  member_name = "firstname-lastname@company.com"
}
```

**Additional context**

I can reproduce the error by running this SQL. Note the quotes are not provided:

```sql
GRANT readonly TO firstname-lastname@company.com;
```

which also results in this same error

```bash
Error: Expected end of statement, found operator "-"
```

aidan-melen:
This is interesting. Looking at the [escape docs](https://materialize.com/docs/sql/types/text/#escape): we tried this and got a little further:

```terraform
resource "materialize_role_grant" "role_grant_user" {
  role_name   = "es_data_ops_readwrite"
  member_name = "firstname-lastname@company\u2024com"
}
```

this time we get this error

```bash
Error: ERROR: unknown role 'firstname-lastname@pluralsight․com' (SQLSTATE XX000)
```

even though the `firstname-lastname@pluralsight․com` does exist.

PR #239, opened 2023-07-15
Fix identifier quoting

benesch:
The `QuoteIdentifier` function was incorrectly splitting identifiers on `.` and then escaping each component. This appears to be cruft left over from an earlier version of the Terraform provider, which was less rigorous in its handling of IDs.

This is a fix for the follow-up report on #233, in which role grants to roles that contained dots (like `joe@materialize.com`) were failing.

### DIFF
--- a/pkg/materialize/utils.go
+++ b/pkg/materialize/utils.go
@@ -4,18 +4,12 @@ import (
 	"strings"
 )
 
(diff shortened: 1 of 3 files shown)


bobbyiliev (approved):
Looks good!

PR #240, opened 2023-07-16
Additional Acc Tests

dehume:
Additional acceptance tests for better coverage for resources where role names are used (all RBAC resources)

https://github.com/MaterializeInc/terraform-provider-materialize/pull/239


benesch (approved):
Awesome, thank you!

PR #245, opened 2023-07-21
Fix Removed Grants

dehume:
If a grant is removed outside of Terraform, it would lead to an error in subsequent plan/applies rather than recreating the grant.

New Workflow
1. Grant via `apply`
```
resource "materialize_table_grant" "table_grant_insert" {
  role_name     = materialize_role.role_1.name
  privilege     = "INSERT"
  database_name = materialize_table.simple_table.database_name
  schema_name   = materialize_table.simple_table.schema_name
  table_name    = materialize_table.simple_table.name
}
```
2. Manually remove grant
```
REVOKE INSERT ON TABLE example_database.example_schema.simple_table FROM "role-1";
```
3. On next plan/apply
```
Terraform will perform the following actions:

  # materialize_table_grant.table_grant_insert will be created
  + resource "materialize_table_grant" "table_grant_insert" {
      + database_name = "example_database"
      + id            = (known after apply)
      + privilege     = "INSERT"
      + role_name     = "role-1"
      + schema_name   = "example_schema"
      + table_name    = "simple_table"
    }

Plan: 1 to add, 0 to change, 0 to destroy.
```

Fix applies to object, system and default grants

### DIFF
--- a/pkg/resources/resource_grant.go
+++ b/pkg/resources/resource_grant.go
@@ -3,6 +3,7 @@ package resources
 import (
 	"context"
 	"fmt"
+	"log"
 	"strings"
 
 	"github.com/MaterializeInc/terraform-provider-materialize/pkg/materialize"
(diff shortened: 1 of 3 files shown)


benesch (commented):
Code LGTM, but we should write a test for this, similar to how we test for objects disappearing:

https://github.com/MaterializeInc/terraform-provider-materialize/blob/7acc978099fa6e264a00597a5e4a1ac0b3e6b0ed/pkg/provider/acceptance_secret_test.go#L74-L98

bobbyiliev (approved):

PR #250, opened 2023-07-26
Adds INCLUDE AS with kafka sources

### LINKED ISSUE #249: Support `INCLUDE KEY AS <name>` for Kafka sources
Kafka sources support aliasing all of the `INCLUDE` options, as in:

  * `INCLUDE KEY AS <name>`
  * `INCLUDE PARTITION AS <name>`
  * `INCLUDE OFFSET AS <name>`
  * `INCLUDE PARTITION AS <name>`
  * `INCLUDE TIMESTAMP AS <name>`

We should support all of these.

Note that we once _required_ specifying an alias to use most of these options, but that got simplified to the current approach of having a single `include_foo` toggle for each option in #100.

One easy way to make this compatible with the current `include_foo` toggles is to introduce new options like:

  * `include_key_alias`
  * `include_partition_alias`
  * `include_offset_alias`
  * `include_partition_alias`
  * `include_timestamp_alias`

And then enforce that you can only set `include_foo_alias` if you've set `include_foo` to true.

The compelling use case for this is an `ENVELOPE UPSERT` source where the key columns and value columns overlap. E.g. a source with columns `id1`, `id2`, `value` where the `id1` and `id2` columns appear in both the source and the value. Attempting to create this source as

```
CREATE SOURCE
FROM KAFKA ...
ENVELOPE UPSERT
```

will fail, because Materialize will attempt to create a source with columns (id1, id2, id1, id2 value), and it's not legal to have repeated columns like that.

Specifying `INCLUDE KEY` doesn't fix the problem, because that's already the default for `ENVELOPE UPSERT`:

```
CREATE SOURCE
FROM KAFKA ...
INCLUDE KEY
ENVELOPE UPSERT
```

The trick is to specify `INCLUDE KEY AS my_key_col`, or somesuch:

```
CREATE SOURCE
FROM KAFKA ...
INCLUDE KEY AS my_key_col
ENVELOPE UPSERT
```

Then the key will be emitted as a _single_ column named `my_key_col`, rather than two columns named `id1` and `id2` that overlap with the value columns of the same name.

bobbyiliev:
Closes #249

### DIFF
--- a/pkg/materialize/source_kafka.go
+++ b/pkg/materialize/source_kafka.go
@@ -333,7 +333,7 @@ func (b *SourceKafkaBuilder) Create() error {
 	var i []string
 
 	if !b.includeKey && b.keyAlias != "" {
-		return fmt.Errorf("key_alias is set but include_key is false")
+		return fmt.Errorf("include_key_alias is set but include_key is false")
 	}
 
 	if b.includeKey {
@@ -357,7 +357,7 @@ func (b *SourceKafkaBuilder) Create() error {
 	}
 
 	if !b.includePartition && b.partitionAlias != "" {
-		return fmt.Errorf("partition_alias is set but include_partition is false")
+		return fmt.Errorf("include_partition_alias is set but include_partition is false")
 	}
 
 	if b.includePartition {
@@ -369,7 +369,7 @@ func (b *SourceKafkaBuilder) Create() error {
 	}
 
 	if !b.includeOffset && b.offsetAlias != "" {
-		return fmt.Errorf("offset_alias is set but include_offset is false")
+		return fmt.Errorf("include_offset_alias is set but include_offset is false")
 	}
 
 	if b.includeOffset {
@@ -381,7 +381,7 @@ func (b *SourceKafkaBuilder) Create() error {
 	}
 
 	if !b.includeTimestamp && b.timestampAlias != "" {
-		return fmt.Errorf("timestamp_alias is set but include_timestamp is false")
+		return fmt.Errorf("include_timestamp_alias is set but include_timestamp is false")
 	}
 
 	if b.includeTimestamp {
--- a/pkg/materialize/source_kafka.go
+++ b/pkg/materialize/source_kafka.go
@@ -38,6 +38,11 @@ type SourceKafkaBuilder struct {
 	includePartition bool
 	includeOffset    bool
 	includeTimestamp bool
+	keyAlias         string
+	headersAlias     string
+	partitionAlias   string
+	offsetAlias      string
+	timestampAlias   string
 	format           SourceFormatSpecStruct
 	keyFormat        SourceFormatSpecStruct
 	valueFormat      SourceFormatSpecStruct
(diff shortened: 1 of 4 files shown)


benesch (commented):
Is there a way to produce an error if you set `include_key = false ` but `include_key_alias = “something”`? Code otherwise looks good!

bobbyiliev:
> Is there a way to produce an error if you set `include_key = false ` but `include_key_alias = “something”`? Code otherwise looks good!

I was also thinking about this today while working on the PR.

What we can probably do is to just add an extra check before each of those, eg:

```go
	if !b.includeHeaders && b.headersAlias != "" {
		return fmt.Errorf("include_headers_alias is set but include_headers is false")
	}
```

benesch:
That’d be great!

pkg/materialize/source_kafka.go
```diff
@@ -297,24 +332,64 @@ func (b *SourceKafkaBuilder) Create() error {
 	// Metadata
 	var i []string
 
+	if !b.includeKey && b.keyAlias != "" {
+		return fmt.Errorf("key_alias is set but include_key is false")
```
benesch, line 336:
```suggestion
		return fmt.Errorf("include_key_alias is set but include_key is false")
```

pkg/materialize/source_kafka.go
```diff
@@ -297,24 +332,64 @@ func (b *SourceKafkaBuilder) Create() error {
...
-		i = append(i, "HEADERS")
+		if b.headersAlias != "" {
+			i = append(i, fmt.Sprintf("HEADERS AS %s", b.headersAlias))
+		} else {
+			i = append(i, "HEADERS")
+		}
+	}
+
+	if !b.includePartition && b.partitionAlias != "" {
+		return fmt.Errorf("partition_alias is set but include_partition is false")
```
benesch, line 360:
```suggestion
		return fmt.Errorf("include_partition_alias is set but include_partition is false")
```

pkg/materialize/source_kafka.go
```diff
@@ -297,24 +332,64 @@ func (b *SourceKafkaBuilder) Create() error {
...
-		i = append(i, "PARTITION")
+		if b.partitionAlias != "" {
+			i = append(i, fmt.Sprintf("PARTITION AS %s", b.partitionAlias))
+		} else {
+			i = append(i, "PARTITION")
+		}
+	}
+
+	if !b.includeOffset && b.offsetAlias != "" {
+		return fmt.Errorf("offset_alias is set but include_offset is false")
```
benesch, line 372:
```suggestion
		return fmt.Errorf("include_offset_alias is set but include_offset is false")
```

pkg/materialize/source_kafka.go
```diff
@@ -297,24 +332,64 @@ func (b *SourceKafkaBuilder) Create() error {
...
-		i = append(i, "OFFSET")
+		if b.offsetAlias != "" {
+			i = append(i, fmt.Sprintf("OFFSET AS %s", b.offsetAlias))
+		} else {
+			i = append(i, "OFFSET")
+		}
+	}
+
+	if !b.includeTimestamp && b.timestampAlias != "" {
+		return fmt.Errorf("timestamp_alias is set but include_timestamp is false")
```
benesch, line 384:
```suggestion
		return fmt.Errorf("include_timestamp_alias is set but include_timestamp is false")
```

benesch (approved):
LGTM. Would be great to also beef up the test to cover more of the cases (`include_key = true` but no `include_key_alias` still works, and `include_key = false` but `include_key_alias` set produces an error), but let's get this into a release in the meantime!

bobbyiliev:
> LGTM. Would be great to also beef up the test to cover more of the cases (`include_key = true` but no `include_key_alias` still works, and `include_key = false` but `include_key_alias` set produces an error), but let's get this into a release in the meantime!

Sounds good! I've created an issue for this so that we don't forget to add those tests: https://github.com/MaterializeInc/terraform-provider-materialize/issues/251

Waiting for the current test run to pass and will prepare the new release.

PR #262, opened 2023-08-01
Support `FOR SCHEMAS` for PostgreSQL Source

dehume:
Support `FOR SCHEMAS` as an alternative for `FOR TABLES` for the Postgres Source

Replaces: https://github.com/MaterializeInc/terraform-provider-materialize/pull/195

### DIFF
--- a/pkg/resources/resource_source_postgres.go
+++ b/pkg/resources/resource_source_postgres.go
@@ -64,9 +64,18 @@ var sourcePostgresSchema = map[string]*schema.Schema{
 				},
 			},
 		},
-		Optional: true,
-		MinItems: 1,
-		ForceNew: true,
+		Optional:      true,
+		MinItems:      1,
+		ForceNew:      true,
+		ConflictsWith: []string{"schema"},
+	},
+	"schema": {
+		Description:   "Creates subsources for specific schemas.",
+		Type:          schema.TypeList,
+		Elem:          &schema.Schema{Type: schema.TypeString},
+		Optional:      true,
+		ForceNew:      true,
+		ConflictsWith: []string{"table"},
 	},
 	"expose_progress": {
 		Description: "The name of the progress subsource for the source. If this is not specified, the subsource will be named `<src_name>_progress`.",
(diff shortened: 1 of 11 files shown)


pkg/resources/resource_source_postgres.go
```diff
@@ -64,9 +64,18 @@ var sourcePostgresSchema = map[string]*schema.Schema{
...
+		ForceNew:      true,
+		ConflictsWith: []string{"schema"},
+	},
+	"schema": {
+		Description:   "Creates subsources for specific schemas.",
+		Type:          schema.TypeList,
+		Elem:          &schema.Schema{Type: schema.TypeString},
+		Optional:      true,
+		ForceNew:      true,
+		ConflictsWith: []string{"table"},
```
morsapaes, line 78: At the time, I was confused about `all_tables` not being explicitly part of the schema. My understanding is that this defaults to adding `ALL TABLES` if neither `table` or `schema` are specified, but we don't seem to do the same for `SIZE` and `IN CLUSTER`.

In both cases, Materialize will fail if you don't type exactly one of the options, so it's confusing to me that in one case we use `ExactlyOneOf`, and in the other we choose an implicit default for a clause that in one of the required options. What am I missing?

dehume: It is a little off. I think the changes that need to happen are:
1. Clusters and Size should use `AtLeastOneOf` instead of `ExactlyOneOf` and include a `ConflictsWith` because we need the user to explicitly provide a value in that case.
```
		AtLeastOneOf:  []string{"cluster_name", "size"},
		ConflictsWith: []string{"size"},
```
2. Using just `ConflictsWith` should be fine for table/schema but both should include an additional line to the description about how if neither is selected, the source will default to `ALL TABLES` which is the case but unclear to the user.

morsapaes:
@dehume, is this good to merge with the code refactoring as a follow-up task?

dehume:
> @dehume, is this good to merge with the code refactoring as a follow-up task?

Yep. Made some changes to line up with https://github.com/MaterializeInc/terraform-provider-materialize/pull/276 and then good to merge in.

bobbyiliev (approved):

Issue #228, opened 2023-07-07
Add support for `ALTER SOURCE...{ADD|DROP} SUBSOURCE`


morsapaes:
Users are now able to remove subsources from a PostgreSQL source ([#20067](https://github.com/MaterializeInc/materialize/pull/20067)), and will soon be able to also add new upstream tables as subsources ([#20277](https://github.com/MaterializeInc/materialize/pull/20277)).

morsapaes:
The final bit of work landed in v0.62, @dehume @bobbyiliev, so this is ready to go.

dehume:
Before we can add this, I think we need source resources to have a better understanding of their subsources https://github.com/MaterializeInc/terraform-provider-materialize/issues/244

PR #263, opened 2023-08-02
Include Subsource Dependencies

dehume:
Include `subsources` as a computed field for source resources. Prep work to allow add/drop of subsources https://github.com/MaterializeInc/terraform-provider-materialize/issues/228.

Debating if this should include additional filtering. Right now every source would include at least one subsource (the progress source). And resources with explicit subsources will contain multiple

```
resource "materialize_source_postgres" "example_source_postgres" {
  name = "source_postgres"
  ...
  table {
    name  = "table1"
    alias = "s1_table1"
  }
  table {
    name  = "table2"
    alias = "s2_table1"
  }
}
```

Would include
```
"subsource": [
    {
    "database_name": "materialize",
    "name": "s1_table1",
    "schema_name": "public"
    },
    {
    "database_name": "materialize",
    "name": "s2_table1",
    "schema_name": "public"
    },
    {
    "database_name": "materialize",
    "name": "source_postgres_progress",
    "schema_name": "public"
    }
]
```
Because we are adding subsources as a computed field, the user would only be able to add/drop subsources via another attribute. This means we could support `ADD|DROP TABLE` like the above postgres source example. I'm still not sure how best to support `ADD|DROP SUBSOURCE`.

### DIFF
--- a/docs/resources/source_kafka.md
+++ b/docs/resources/source_kafka.md
@@ -81,6 +81,7 @@ resource "materialize_source_kafka" "example_source_kafka" {
 
 - `id` (String) The ID of this resource.
 - `qualified_sql_name` (String) The fully qualified name of the source.
+- `subsource` (List of Object) Subsources of a source. (see [below for nested schema](#nestedatt--subsource))
 
 <a id="nestedblock--kafka_connection"></a>
 ### Nested Schema for `kafka_connection`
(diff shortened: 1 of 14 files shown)


benesch:
Won't have time to review in depth, but skimmed and it checked out!

bobbyiliev (approved):
Looks good!

PR #265, opened 2023-08-04
Support Add/Drop Subsource

dehume:
Support `ALTER SOURCE...{ADD|DROP} SUBSOURCE` for postgres source resources.

https://materialize.com/docs/sql/alter-source/#add-subsource-with_options

This is separate from the `subsource` attribute which is a computed field. Adding our removing from the `tables` attribute will trigger a corresponding `ADD|DROP` of the subsource. If the user is adding a table and modifying the `text_columns` attribute, the difference between the current state of the text columns and the new text columns will be added to the `ADD SUBSOURCE` command.

Currently this is only applied to the postgres source. Though it does apply to the load generator source for the auction house example (though seems less likely of a use case in production managed with TF).


dehume:
NOTE. This implementation would not work for `ALL TABLES` the table declarations would need to be explicit currently.

integration/postgres/postgres_bootstrap.sql
```diff
@@ -10,12 +10,18 @@ CREATE TABLE table2 (
...
 ALTER TABLE table2 REPLICA IDENTITY FULL;
+ALTER TABLE table3 REPLICA IDENTITY FULL;
 
 -- Create publication on the created tables
-CREATE PUBLICATION mz_source FOR TABLE table1, table2;
+CREATE PUBLICATION mz_source FOR TABLE table1, table2, table3;
 
 INSERT INTO table1 VALUES (1), (2), (3), (4), (5);
 INSERT INTO table2 VALUES (1, NOW()), (2, NOW()), (3, NOW()), (4, NOW()), (5, NOW());
+INSERT INTO table3 VALUES (1), (2), (3), (4), (5);
```
sploiselle, line 27: nbd but this looks like it's missing a trailing newline

pkg/materialize/source.go
```diff
@@ -23,6 +24,44 @@ func GetTableStruct(v []interface{}) []TableStruct {
 	return tables
 }
 
+func DiffTableStructs(arr1, arr2 []interface{}) []TableStruct {
+	var difference []TableStruct
+
+	for _, item1 := range arr1 {
+		found := false
```
sploiselle, line 31: `found` is redundant here because the only case  where you enter the second block is when you haven't encountered a `break`

pkg/resources/resource_source_postgres_test.go
```diff
@@ -52,3 +56,41 @@ func TestResourceSourcePostgresCreate(t *testing.T) {
...
+	d := schema.TestResourceDataRaw(t, SourcePostgres().Schema, inSourcePostgres)
+
+	d.SetId("u1")
+	d.Set("size", "large")
+	r.NotNil(d)
+
+	testhelpers.WithMockDb(t, func(db *sqlx.DB, mock sqlmock.Sqlmock) {
+		mock.ExpectExec(`ALTER SOURCE "database"."schema"."source" SET \(SIZE = 'small'\)`).WillReturnResult(sqlmock.NewResult(1, 1))
+		mock.ExpectExec(`ALTER SOURCE "database"."schema"."" RENAME TO "source"`).WillReturnResult(sqlmock.NewResult(1, 1))
+		mock.ExpectExec(`ALTER SOURCE "database"."schema"."source" ADD SUBSOURCE "name1" AS "alias", "name2"`).WillReturnResult(sqlmock.NewResult(1, 1))
```
sploiselle, line 71: Doesn't look like there's a test for `DROP SUBSOURCE`.

sploiselle (commented):
I'm wholly unfamiliar with the code in this repo, so can you point me to the test that changes the resource definition of the existing source and demos that the statements we generate to move between states is accurate? Just need some help parsing out everything happening here. Thanks!

dehume:
> I'm wholly unfamiliar with the code in this repo, so can you point me to the test that changes the resource definition of the existing source and demos that the statements we generate to move between states is accurate? Just need some help parsing out everything happening here. Thanks!

The tests for the provider are tough, especially updates. The unit tests for the resource are minimal because you have to hack TF state which can be finicky. The more through tests are in `pkg/provider/acceptance_source_postgres_test.go` which actually run `terraform apply` against the materialized image.

It can be a little tough to read but the acceptance tests for updates are running through 3 steps (three different `terraform apply` runs). 

1. Create the initial source (leaving out parts of the TF and SQL that don't have to do with tables or text columns)
```
resource "materialize_source_postgres" "test" {
        ...
	table {
		name  = "table1"
		alias = "conn_table1"
	}
	table {
		name  = "table2"
		alias = "conn_table2"
	}
	text_columns = ["table1.id"]
}
```
Which is equivalent to roughly:
```
CREATE SOURCE ... (TEXT COLUMNS (table1.id)) FOR TABLES (table1 AS conn_table1, table2 AS conn_table2)
```
2. Run the update `terraform apply` by changing the `table` and `text_column` attributes.
```
resource "materialize_source_postgres" "test" {
        ...
	table {
		name  = "table1"
		alias = "conn_table1"
	}
	table {
		name  = "table3"
		alias = "conn_table3"
	}
	text_columns = ["table1.id", "table3.id"]
}
```
Which is equivalent to:
```
ALTER SOURCE ... ADD SUBSOURCE "table_3" AS "conn_table3" WITH (TEXT COLUMNS [table3.id]);
ALTER SOURCE ... DROP SUBSOURCE "conn_table2";
```
3. Reapply the first terraform config. Which is equivalent to:
```
ALTER SOURCE ... ADD SUBSOURCE "table_2" AS "conn_table2";
ALTER SOURCE ... DROP SUBSOURCE "conn_table3";
```

pkg/materialize/source.go
dehume: I tried removing it but getting different results. What were you thinking in how it should be written?

pkg/resources/resource_source_postgres_test.go
dehume: Mentioned in more detail below, but the unit tests for updates are hard and kept to a minimum for the most part.

pkg/materialize/source.go
```diff
@@ -23,6 +24,44 @@ func GetTableStruct(v []interface{}) []TableStruct {
...
+			}
+		}
+		if !found {
+			if diffItem, ok := item1.(map[string]interface{}); ok {
+				difference = append(difference, TableStruct{
+					Name:  diffItem["name"].(string),
+					Alias: diffItem["alias"].(string),
+				})
+			}
+		}
```
sploiselle, line 45: To remove `found`, you need to break out of the outer loop, not just let the inner loop complete.
```suggestion
out:
	for _, item1 := range arr1 {
		for _, item2 := range arr2 {
			if areEqual(item1, item2) {
				break out
			}
		}
		if diffItem, ok := item1.(map[string]interface{}); ok {
			difference = append(difference, TableStruct{
				Name:  diffItem["name"].(string),
				Alias: diffItem["alias"].(string),
			})
		}
	}
```

sploiselle (approved):
tysm for the explanation--these changes make sense to me, though probably need someone more familiar with the package's idioms to also review

PR #271, opened 2023-08-09
Initial webhook source setup

### LINKED ISSUE #270: Resource `Webhook` Source
Include support for webhook source type.

https://materialize.com/docs/sql/create-source/webhook/

bobbyiliev:
Adding the very initial work for supporting webhook sources.

Things that still need to be ironed out:

- [x] Add more acceptance tests
- [ ] Improve `checkExpression`
- [x] Improve `checkOptions`
- [x] I believe that `size` does not apply to the webhook source, at least judging by the [grammar file](https://github.com/MaterializeInc/materialize/blob/main/doc/user/sql-grammar/sql-grammar.bnf#L158-L170)
- [x] Integration tests

Closes #270


pkg/materialize/source_webhook.go
```diff
@@ -0,0 +1,82 @@
...
+	"github.com/jmoiron/sqlx"
+)
+
+type SourceWebhookBuilder struct {
+	Source
+	clusterName     string
+	size            string // TODO: size is not supported for webhook sources
+	bodyFormat      string
+	includeHeaders  bool
+	checkOptions    []string
```
dehume, line 16: Should this be a custom struct like
```
type checkOptions struct {
    field string
    alias string
}
```
Then in the resource we can put in validation to ensure they are only using a field of type `BODY`, `HEADER`, or `SECRET`

pkg/provider/acceptance_source_webhook_test.go
```diff
@@ -0,0 +1,88 @@
...
+resource "materialize_source_webhook" "test" {
+	name = "%[1]s"
+	cluster_name = materialize_cluster.example_cluster.name
+	body_format = "json"
+	include_headers = false
+}
+`, sourceName)
+}
+
+func testAccCheckSourceWebhookExists(name string) resource.TestCheckFunc {
```
dehume, line 59: Realizing we could just use the same function for all source acceptance tests rather than defining a custom one for each source type. At least for any source that just uses `ScanSource`.

pkg/resources/resource_source_webhook.go
```diff
@@ -0,0 +1,115 @@
...
+		checkOptionsStrings[i] = option.(string)
+	}
+
+	b.CheckOptions(checkOptionsStrings)
+
+	// Create resource
+	if err := b.Create(); err != nil {
+		return diag.FromErr(err)
+	}
+
```
dehume, line 106: Since `ownership` can be set. Will need a [block for the adding ownership](https://github.com/MaterializeInc/terraform-provider-materialize/blob/main/pkg/resources/resource_source_kafka.go#L282-L291).

bobbyiliev: Good point! Just added this.

pkg/materialize/source_webhook.go
bobbyiliev: Just updated this, but also realized that it might be a bit more complex than my current implementation as I'll have to add a check for the SECRET as well to make sure that an actual secret is provided

dehume (approved):

PR #275, opened 2023-08-11
[WIP] Managed Cluster Options

dehume:
Support for the additional attributes that can be set with managed clusters:
* attributes that had been specific to cluster replicas
* All for additional `ALTER` in update
* Managed clusters can be set with just `size` (replication factor not necessary)


dehume:
Leaving out availability zone for now

bobbyiliev (approved):
Looks good!

PR #298, opened 2023-08-30
Object Mapping

dehume:
Rename `ObjectSchemaStruct` to `MaterializeObject` and be more consistent with initializing the `MaterializeObject` and builder which shows up most often in the update context functions were was inconsistently being redeclared.


bobbyiliev (approved):
LGTM!

PR #300, opened 2023-08-31
Default Grants Support `All Roles`

dehume:
All for `All Roles`. To better align with materialize, to set this properly the TF user will use `PUBLIC` which is a reserved word within roles.

NOTE: This is more of a hotfix to unblock FE. Occasionally seeing some oddities on destroy with more complicated TF deployments


bobbyiliev (approved):
Looks good!

PR #310, opened 2023-09-11
Require `FOR ALL TABLES` Multi Output Sources

dehume:
Include `FOR ALL TABLES` as a non configurable part of the SQL syntax for multi output load generator sources (auction and marketing)


bobbyiliev (approved):
LGTM!

PR #311, opened 2023-09-11
docs(source): update postgres source example

ahstn:
This PR contains the following:
* Minor update to reflect `table` being a `block` not an `object` for `materialize_source_postgres` (see error below)
  * Current docs: [materialize_source_postgres | Terraform](https://registry.terraform.io/providers/MaterializeInc/materialize/latest/docs/resources/source_postgres)
  * Correctly ran `make docs` this time 😅 
* Typo correction in `postgres_connection`: "The posgres_connection name."
* Adding guidance on finding `materialize_connection_postgres` ID in Materialize
* Additional `materialize_connection_postgres` to show non-secret username and SSH tunnel config

```
➜ tf plan
╷
│ Error: Unsupported argument
│
│   on main.tf line 13, in resource "materialize_source_postgres" "this":
│   13:   table = {
│
│ An argument named "table" is not expected here. Did you mean to define a block of type "table"?
```


examples/resources/materialize_connection_postgres/import.sh
```diff
@@ -1,2 +1,6 @@
-#Connections can be imported using the connection id:
+# Connections can be imported using the `connection id`:
 terraform import materialize_connection_postgres.example <connection_id>
+
+# You can find the `connection_id` from the following query:
```
dehume, line 4: Can we leave this out of the example for now? If we are going to include these queries to lookup the resource id  for imports, we should probably include them for all examples in the directory

ahstn: Yup can do 👍 . Was just a bit confusing initially as I was trying the "name" thinking it was the identifier.

dehume (approved):

bobbyiliev (approved):
Thank you for this contribution!

bobbyiliev (changes requested):
Just noticed that it looks like the docs workflow is failing as the `docs/resources/connection_postgres.md` file was changed in the PR. That file is auto-generated based on the contents of `examples/resources/materialize_connection_postgres/resource.tf` so there is no need to edit the `connection_postgres.md` directly or commit it.

dehume:
> Just noticed that it looks like the docs workflow is failing as the `docs/resources/connection_postgres.md` file was changed in the PR.

I wrote up an [issue](https://github.com/MaterializeInc/terraform-provider-materialize/issues/313) for this. I think there is something about the Github Action that fails on subsequent pushes to branches on forks. I've noticed this once before on another PR.

Short term solution is that this can just be merged and whatever the next PR that is put up will generate the correct documentation.

ahstn:
Ah okay, I can revert the `docs/` changes tomorrow then hopefully the GHA workflow doesn't conflict

dehume:
Yeah I don't think it has to do anything with @ahstn 's PR so we can probably merge this and figure out the GHA tomorrow or something.

bobbyiliev (approved):

bobbyiliev:
Sounds good! Thank you again for the contribution @ahstn!

PR #314, opened 2023-09-11
Postgres `schema` Attribute

dehume:
Using the `schema` attribute within the postgres source was leading to errors. Including fix and including unit and integration tests specific for schema attribute.

### DIFF
--- a/integration/source.tf
+++ b/integration/source.tf
@@ -42,6 +42,18 @@ resource "materialize_source_postgres" "example_source_postgres" {
   text_columns = ["table1.id"]
 }
 
+resource "materialize_source_postgres" "example_source_postgres_schema" {
+  name = "source_postgres_schema"
+  size = "2"
+  postgres_connection {
+    name          = materialize_connection_postgres.postgres_connection.name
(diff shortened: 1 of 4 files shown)


integration/source.tf
```diff
@@ -42,6 +42,18 @@ resource "materialize_source_postgres" "example_source_postgres" {
...
+resource "materialize_source_postgres" "example_source_postgres_schema" {
+  name = "source_postgres_schema"
+  size = "2"
+  postgres_connection {
+    name          = materialize_connection_postgres.postgres_connection.name
+    schema_name   = materialize_connection_postgres.postgres_connection.schema_name
+    database_name = materialize_connection_postgres.postgres_connection.database_name
+  }
+  publication = "mz_source"
+  schema = ["PUBLIC"]
```
bobbyiliev, line 54: Seems like the terraform fmt is failing bacause of this
```suggestion
  publication = "mz_source"
  schema      = ["PUBLIC"]
```

bobbyiliev (approved):

PR #315, opened 2023-09-12
GHA Docs Forks

dehume:
When users submit PRs on forks, the documentation check will fail. This is because of how the [git-auto-commit-action works with public repos](https://github.com/stefanzweifel/git-auto-commit-action#use-in-forks-from-public-repositories). Seems like the correct (hopefully simple) solution is to changing the trigger


bobbyiliev (approved):
Neat! Looks good

PR #324, opened 2023-09-22
Attribute `COMMENT`

dehume:
Add support for [`COMMENT`](https://materialize.com/docs/sql/comment-on/) for both objects and columns within objects:

- [x] Cluster
- [x] Cluster Replica
- [x] Database
- [x] Index (Materialize Object issue)
- [x] Materialized View
- [x] Role (Materialize Object issue)
- [x] Schema
- [x] Secret
- [x] Sink
- [x] Source (Multi)
- [x] Connection (Multi)
- [x] Table
- [x] Table (Column)
- [x] Type
- [x] View

### DIFF
--- a/pkg/resources/resource_cluster_replica.go
+++ b/pkg/resources/resource_cluster_replica.go
@@ -3,6 +3,7 @@ package resources
 import (
 	"context"
 	"database/sql"
+	"log"
 
 	"github.com/MaterializeInc/terraform-provider-materialize/pkg/materialize"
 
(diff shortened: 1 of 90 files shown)


dehume:
Using Table as a POC, will then implement with other objects

pkg/resources/resource_cluster_replica.go
```diff
@@ -14,6 +15,7 @@ import (
 var clusterReplicaSchema = map[string]*schema.Schema{
 	"name":         ObjectNameSchema("replica", true, true),
 	"cluster_name": ClusterNameSchema(),
+	"comment":      CommentSchema(true),
```
dehume, line 18: Three resources: roles, cluster replicas and indexes are still marked as force new. Breaking out into a followup issue https://github.com/MaterializeInc/terraform-provider-materialize/issues/329

bobbyiliev (approved):
Looks good! Thank you for taking care of all those changes!

PR #325, opened 2023-09-22
Remove conditional logic for enabled flags

morsapaes:
Both webhook sources and connection validation are now in public preview, so we no longer need to enable them for testing.

### DIFF
--- a/pkg/provider/provider.go
+++ b/pkg/provider/provider.go
@@ -169,31 +169,5 @@ func providerConfigure(ctx context.Context, d *schema.ResourceData) (interface{}
 		return nil, diags
 	}
 
-	if testing {
-		// TODO: Remove this once enable_webhook_sources is enabled by default
-		_, err = db.Exec("ALTER SYSTEM SET enable_webhook_sources = true;")
-		if err != nil {
-			diags = append(diags, diag.Diagnostic{
-				Severity: diag.Error,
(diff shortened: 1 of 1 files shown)


pkg/provider/provider.go
```diff
@@ -169,31 +169,5 @@ func providerConfigure(ctx context.Context, d *schema.ResourceData) (interface{}
...
-				Summary:  "Unable to enable webhook sources",
-				Detail:   "Unable to enable webhook sources for authenticated Materialize client",
-			})
-			return nil, diags
-		}
-
-		// TODO: Remove this once enable_connection_validation_syntax is enabled by default
-		_, err = db.Exec("ALTER SYSTEM SET enable_connection_validation_syntax = true;")
-		// TODO: Remove this once enable_disk_cluster_replicas is enabled by default
-		_, err = db.Exec("ALTER SYSTEM SET enable_disk_cluster_replicas TO true;")
```
morsapaes, line 187: Oh, I missed `enable_disk_cluster_replicas`. This is still gated. AFK now but will patch once I'm back on if no one gets to it first!

dehume:
I think we just need to wait for them to be supported in the image so the integration tests will pass. But I can clean up the `**Private Preview**` in https://github.com/MaterializeInc/terraform-provider-materialize/pull/323

PR #330, opened 2023-10-06
Role Session Variable

dehume:
Allow session variables to be included as part of `Role` resources.


pkg/provider/provider.go
```diff
@@ -171,48 +171,25 @@ func providerConfigure(ctx context.Context, d *schema.ResourceData) (interface{}
...
-		// TODO: Remove this once enable_webhook_sources is enabled by default
-		_, err = db.Exec("ALTER SYSTEM SET enable_webhook_sources = true;")
-		if err != nil {
-			diags = append(diags, diag.Diagnostic{
-				Severity: diag.Error,
-				Summary:  "Unable to enable webhook sources",
-				Detail:   "Unable to enable webhook sources for authenticated Materialize client",
-			})
-			return nil, diags
+		flags := []string{
```
dehume, line 174: Simplifying feature flags

pkg/resources/config.go
```diff
@@ -79,3 +79,35 @@ var aliases = map[string]string{
 	"int2":    "smallint",
 	"uint":    "uint4",
 }
+
+var session_variables = []string{
```
dehume, line 83: This is just used for validation within the resource. Not sure if we want to this just since it might be harder to keep things in sync.

dehume (commented):
This allows role session variables to be set upon creation. However there is not a way (that I know of) to query the state of these session variables after creation.

bobbyiliev (approved):

PR #335, opened 2023-10-19
Mask Local Sizes

dehume:
Mask the local sizes for cluster replicas used by the docker image so these options will not be displayed in the validation for the resource attributes.

Limiting to only allowing `3xsmall` and `2xsmall` (though both map to 1).


bobbyiliev (approved):
Nice!

PR #339, opened 2023-10-24
178 Provider configuration params unification

### LINKED ISSUE #178: Provider configuration thoughts
The provider currently has some configuration parameters that are spelled differently depending on whether you specify them in a `provider` block in your `tf` file or as environment variables:

  * `username` vs `MZ_USER`
  * `password` vs `MZ_PW`

By convention I believe these are all meant to be identical, so `MZ_USERNAME` and `MZ_PASSWORD`.

We should also see whether there's a convention around always using the full provider name as the environment variable prefix. E.g., should we be looking at `MATERIALIZE_HOST` instead of `MZ_HOST`?

I also think the `testing`/`MZ_TESTING` parameter is a bit of a smell. Can we rename that to `MZ_SSL_MODE` or `MZ_SSLMODE` and accept `disabled` and `required` as options? We could leave it undocumented since cloud users will always be using TLS, but since the `testing` flag _only_ controls the SSL mode I think it's best to just name the option after that.

Note also that the "[Terraform Cloud Resources](https://www.notion.so/materialize/Terraform-Cloud-Resources-839b26be77b5493b8ad319f02769f4d2?pvs=4#54a7afd2f62a49d9891fb9ccf79b2ce2)" doc might result in some other big changes to the provider configuration—i.e., the possible introduction of `app_password`, `default_region`, and `endpoint` fields.

bobbyiliev:
As discussed in #178 updating the provider configuration parameters so that they are consistent across all components.

Closes #178

This is going to be a breaking change for anyone currently using the provider and they will need to update their configurations accordingly, so we should plan this before merging and communicate it accordingly.

On another note on whether we should use `materialize_` vs `mz_` there isn't anything specific about this in the Terraform plugins best practices documentation [here](https://developer.hashicorp.com/terraform/plugin/best-practices). So totally open to suggestions here!

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,7 +29,7 @@ Configure the provider by adding the following block to your Terraform project:
 ```hcl
 provider "materialize" {
   host     = "materialized_hostname"
-  username = "materialize_username"
+  user     = "materialize_user"
   password = "materialize_password"
   port     = 6875
   database = "materialize"


pkg/provider/provider.go
```diff
@@ -126,18 +126,18 @@ func Provider() *schema.Provider {
...
+func connectionString(host, username, password string, port int, database string, sslmode bool, application string) string {
 	c := strings.Builder{}
 	c.WriteString(fmt.Sprintf("postgres://%s:%s@%s:%d/%s", username, password, host, port, database))
 
 	params := []string{}
 
-	if testing {
+	if sslmode {
+		params = append(params, "sslmode=require")
+	} else {
```
dehume, line 137: Just realizing that we are setting feature flags twice, so we can probably remove
```
		params = append(params, "enable_rbac_checks=true")
		params = append(params, "enable_ld_rbac_checks=true")
```
Since all feature flags are set further down. Bigger discussion point I think this should still be called `testing` or something like that rather than `mz_sslmode` since setting it disables ssl and sets a number of specific feature flags.

Or maybe we break this out and have two parameters (hidden from users) that are `mz_sslmode` and `mz_preview_features` which is used to enable the specific feature flags.

dehume (commented):
Overall I think this looks good. We should also figure out any [documentation](https://materialize.com/docs/manage/terraform/) or other repos we would need to update.

This could probably be the first release for `0.2.0`. Worth getting Nikhil's thoughts on rollout and giving users a heads up.

pkg/provider/provider.go
bobbyiliev: I like the idea of breaking this into two parameters! Will work on this today

pkg/provider/provider.go
```diff
@@ -170,7 +175,7 @@ func providerConfigure(ctx context.Context, d *schema.ResourceData) (interface{}
 	}
 
 	// Feature flags to enable
-	if testing {
+	if preview_features {
```
benesch, line 178: Just seeing this block now! I don't think you want this to be part of the TF provider itself? I think you just want the test script to enable these things:

```
environmentd \
  --system-parameter-default=enable_webhook_sources=true \
  --system-parameter-default=enable_connection_validation_syntax=true
```

You can also just use the `--all-features` flag if you'd like to unconditionally enable all feature flags, without manually managing the list.

benesch (changes requested):

pkg/provider/provider.go
```diff
@@ -17,32 +17,32 @@ import (
 func Provider() *schema.Provider {
 	return &schema.Provider{
 		Schema: map[string]*schema.Schema{
-			"host": {
+			"mz_host": {
```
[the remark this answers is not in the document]
bobbyiliev: Ah yes! very good point, it was my bad incorrectly interpreting the issue requirements. Just reverted that back accordingly.

pkg/provider/provider.go
bobbyiliev: Oh neat! Just discussed this with @dehume and updated it accordingly.

pkg/provider/provider.go
```diff
@@ -26,15 +26,15 @@ func Provider() *schema.Provider {
 			"username": {
 				Type:        schema.TypeString,
 				Optional:    true,
-				Description: "Materialize username. Can also come from the `MZ_USER` environment variable.",
-				DefaultFunc: schema.EnvDefaultFunc("MZ_USER", nil),
+				Description: "Materialize username. Can also come from the `MZ_USERNAME` environment variable.",
```
benesch, line 29: Should this one be `user` and `MZ_USER`? PostgreSQL calls this parameter "user" rather than username, at least.

.github/workflows/acceptance.yml
```diff
@@ -18,10 +18,11 @@ jobs:
...
         run: make testacc
         env:
-            MZ_HOST: localhost
-            MZ_USER: mz_system
-            MZ_TESTING: "true"
-            MZ_PORT: 6877
+          MZ_HOST: localhost
+          MZ_USERNAME: mz_system
+          MZ_SSLMODE: "false"
+          MZ_PREVIEW_FEATURES: "true"
```
benesch, line 24:
```suggestion
```

CONTRIBUTING.md
```diff
@@ -41,8 +41,9 @@ To run the acceptance tests which will simulate running Terraform commands you w
 
 ```bash
 export MZ_HOST=localhost
-export MZ_USER=mz_system
-export MZ_TESTING="true"
+export MZ_USERNAME=mz_system
+export MZ_SSLMODE="false"
+export MZ_PREVIEW_FEATURES="true"
```
benesch, line 46:
```suggestion
```

benesch (approved):
This looks great, modulo one comment about `user` vs `username`!

pkg/provider/provider.go
bobbyiliev: Neat! +1 for `user` from me!
Tagging @dehume to see what his thoughts are as well

benesch (approved):
Fully LGTM! Thanks so much for this, @bobbyiliev. I'm 👍🏽 on shipping this as v0.2 and helping our existing customers through the migration.

pkg/provider/provider.go
dehume: `user` and `MZ_USER` sounds good

dehume (approved):

PR #341, opened 2023-10-27
Allow Null Assertions

dehume:
Allow [non null assertions](https://materialize.com/docs/sql/create-materialized-view/#non-null-assertions) in materialized views resource.


bobbyiliev (approved):
Looks good!

PR #342, opened 2023-10-30
Update changelog for v0.2.0

bobbyiliev:


CHANGELOG.md
```diff
@@ -1,5 +1,16 @@
 # Changelog
 
+## 0.2.0 - 2023-10-30
+
+### Features
```
dehume, line 5: I would just use a "Breaking Changes" header. We don't always include it but when there are breaking changes, that is listed as the fourth header

bobbyiliev: Good idea!

dehume (approved):

PR #346, opened 2023-11-03
Update Webhook Source

dehume:
Including additional functionality for [webhook source](https://materialize.com/docs/sql/create-source/webhook/). Specifically:

* Adding the ability to [expose headers](https://materialize.com/docs/sql/create-source/webhook/#exposing-headers) with `INCLUDE HEADER` via the `include_header` attribute
* Ability to [exclude headers](https://materialize.com/docs/sql/create-source/webhook/#excluding-header-fields). This is now back of `include_headers` which allows users to specifically set headers using `only`, excluding with `not` and just setting `INCLUDE HEADERS` by setting `all` to true.
```
	include_headers {
                all    = true
		only = ["a", "b"] -- cannot be used with all
		not  = ["c", "d"] -- cannot be used with all 
	}
```
* Ability to set check options as `BYTES`

Also adding more use cases from the [documentation](https://materialize.com/docs/sql/create-source/webhook/#exposing-headers) in the unit tests as well as the specific use cases for [segment](https://materialize.com/docs/ingest-data/segment/) and [rudderstack](https://materialize.com/docs/ingest-data/rudderstack/).

### DIFF
--- a/docs/resources/source_webhook.md
+++ b/docs/resources/source_webhook.md
@@ -36,6 +36,8 @@ resource "materialize_source_webhook" "example_webhook" {
 - `cluster_name` (String) The cluster to maintain this source.
 - `comment` (String) **Private Preview** Comment on an object in the database.
 - `database_name` (String) The identifier for the source database. Defaults to `MZ_DATABASE` environment variable if set or `materialize` if environment variable is not set.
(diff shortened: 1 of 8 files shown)


bobbyiliev (approved):

bobbyiliev (approved):
Looks good! Thank you for working on this!

PR #349, opened 2023-11-07
Update resource.tf

hui-zheng:
fix the example for terraform.materialize_connection_kafka
before this fix, it throws an error 
```

│   on main.tf line 45, in resource "materialize_connection_kafka" "upstash":
│   45:   sasl_username = {
│
│ An argument named "sasl_username" is not expected here. Did you mean to define a block of type "sasl_username"?
╵
```


bobbyiliev (approved):
Thank you for taking the time and submitting this fix! Looks good!

dehume (approved):
Thank you for contributing. The one Github Action change is on our end and we will sort it out and it will be part of the next release this week.

PR #351, opened 2023-11-08
Cloud Resources Refactor

bobbyiliev:
I've been thinking with #5 based on the doc [here](https://www.notion.so/materialize/Terraform-Cloud-Resources-839b26be77b5493b8ad319f02769f4d2)

Each resource within our provider will be able to utilize one of these clients depending on its requirements. For example, a resource that manages database configurations will use the `DBClient`, whereas another that interacts with Frontegg services will use `FronteggClient` and etc. 

```go
struct {
	DB       *clients.DBClient
	Frontegg *clients.FronteggClient
}
```

This PR is more of a PoC so we could start a further discussion, it implements a `materialize_app_password` resource that utilizes the FrontEgg API:

```sql
resource "materialize_app_password" "example_password" {
  name = "tf-example-app-password2"
}
```

List of things that need to be done to get this PR merged:

- [x] Add a design doc
- [x] Add Frontegg Client
- [x] Add Tests for the Frontegg Client
- [x] Add Cloud API Client
- [x] Add Database Client
- [x] Add Cloud API Resources
- [x] Add Tests for the Cloud API Client
- [x] Refactor the Provider configuration to use the new clients
- [x] Refactor all existing database resources to use the client
- [x] Refactor all existing tests to use the client
- [x] Add the new Cloud Resources:
    - [x] `materialize_app_password`
    - [x] `materialize_user`
- [x] Add the new data sources:
    - [x] `materialize_regions`

Refactor resources to use the new DB client:

- [x] Cluster Replicas
- [x] Clusters
- [x] Connection - AWS Privatelink
- [x] Connection - Confluent Schema Registry
- [x] Connection - Kafka
- [x] Connection - Postgres
- [x] Connection - SSH
- [x] Databases
- [x] Grants
- [x] Grants Default
- [x] Roles
- [x] Indexes
- [x] Materialized Views
- [x] Schemas
- [x] Secrets
- [x] Sinks - Kafka
- [x] Sources - Kafka
- [x] Sources - Load Generator
- [x] Sources - Postgres
- [x] Sources - Webhook
- [x] Tables
- [x] Types
- [x] Views

Tasks that can be contributed separately in follow-up PRs:
- [ ] Add the new Cloud Resources:
    - [ ] `materialize_region`
- [ ] Add the new data sources:
    - [ ] `materialize_organization`


benesch:
Nice, this makes sense to me! Note that you probably want to support multiple "DB" clients, one per region:

```
type Region string

const (
        awsUsEast1 Region = "aws/us-east-1"
        awsUsWest2 Region = "aws/us-west-2"
)

struct {
	DB       map[Region]*clients.DBClient
	Frontegg *clients.FronteggClient
}
```

pkg/provider/provider.go
```diff
@@ -60,8 +61,16 @@ func Provider() *schema.Provider {
 				DefaultFunc: schema.EnvDefaultFunc("MZ_SSLMODE", true),
 				Description: "For testing purposes, disable SSL.",
 			},
+			"endpoint": {
```
dehume, line 64: Is the admin endpoint unique to each organization's Materialize account?

dehume (commented):
This looks great. Exciting to add these

pkg/provider/provider.go
benesch: No, but you'll need some way to override it for tests. Production and staging have different admin endpoints, for example.

docs/developer/design-v2-2023-11-11.md
```diff
@@ -0,0 +1,198 @@
...
+
+### Provider Configuration
+
+The provider will be initialized with parameters that determine its behavior and default settings:
+
+```hcl
+provider "materialize" {
+  app_password   = var.materialize_app_password
+  default_region = var.materialize_default_region
+  endpoint       = var.materialize_endpoint
```
benesch, line 29: We should hide this one from users.

docs/developer/design-v2-2023-11-11.md
```diff
@@ -0,0 +1,198 @@
...
+
+### Multi-Client Structure
+
+The provider will maintain a map of clients for database operations across different regions and a client for interfacing with the Frontegg API.
+
+```hcl
+type ProviderMeta struct {
+    DB       map[Region]*clients.DBClient
+    Frontegg *clients.FronteggClient
+    // Future: CloudAPI *clients.CloudAPIClient
```
benesch, line 43: 👍🏽

docs/developer/design-v2-2023-11-11.md
```diff
@@ -0,0 +1,198 @@
...
+-   **Enable Region**: Uses the Cloud API to enable a new region.
+-   **Read Region**: Retrieves the status of a region from the Cloud API.
+-   **Disable Region**: Noop. Regions cannot be disabled once enabled.
+
+### User Resource
+
+Manages the lifecycle of a user within the Frontegg platform.
+
+```hcl
+resource "materialize_app_user" "example_user" {
```
benesch, line 72: The "app" in "app password" is for "application-specific password"; for the other organization-scoped objects (vs region scoped objects) we don't want to use "app" in the name!

```suggestion
resource "materialize_user" "example_user" {
```

docs/developer/design-v2-2023-11-11.md
```diff
@@ -0,0 +1,198 @@
...
+-   **Read User**: Retrieves user details from Frontegg.
+-   **Update User**: Attaches or detaches roles from the user.
+-   **Delete User**: Removes the user from Frontegg.
+
+### Organization Data Source?
+
+Generates a list of available organizations from Frontegg.
+
+```hcl
+data "materialize_app_organizations" "all" {}
```
benesch, line 92: App passwords are specific to a single organization, so if we want a data source here, it should just be for the single organization that the app password is tied to.

docs/developer/design-v2-2023-11-11.md
```diff
@@ -0,0 +1,198 @@
...
+Operations:
+
+-   **List Organizations**: Fetches all available organizations from Frontegg.
+
+### Role Data Source
+
+Generates a list of available roles from Frontegg.
+
+```hcl
+data "materialize_app_roles" "all" {}
```
benesch, line 104:
```suggestion
data "materialize_organization_roles" "all" {}
```

We need a term for this that distinguishes it from region roles that's not "app". I propose "organization roles"! This fits well with the existing naming ("Organization Admin", "Organization Member").

docs/developer/design-v2-2023-11-11.md
```diff
@@ -0,0 +1,198 @@
...
+
+### User Resource
+
+Manages the lifecycle of a user within the Frontegg platform.
+
+```hcl
+resource "materialize_app_user" "example_user" {
+  email          = "user@example.com"
+  auth_provider  = "local"
+  role_ids       = ["role-id-1", "role-id-2"] # Logic to retrieve role IDs from role names?
```
benesch, line 75:
```suggestion
  organization_roles  = ["admin", "member"]
```

Writing the logic to map these organization role names to their IDs in the provider would be great!

docs/developer/design-v2-2023-11-11.md
```diff
@@ -0,0 +1,198 @@
...
+
+-   **List Roles**: Fetches all available roles from Frontegg.
+
+## Testing Strategy
+
+Along the existing unit tests described in the [design document](../../DESIGN.md), the provider will include a comprehensive suite of tests for all resources and operations, including:
+
+-   **Unit Tests**: Will run against a test Materialize environment or a mock service to ensure that the provider functions correctly in isolation.
+-   **Integration Tests**: Will run against staging instance to confirm that the provider interacts correctly with the actual services.
+-   **Acceptance Tests**: То be determined.
```
benesch, line 117: One thing to be careful of here: I think CI currently runs integration tests against a database Docker container. Once you start using the cloud API, you're not going to be able to use the Docker image alone to run integration tests, because that database Docker container won't have the cloud APIs.

There are a few options:

  * @mjibson is working on a mock Frontegg implementation you could use to mock out Frontegg. (That won't help you with the region enable API though; you'd have to write your own mock for that.)
  * You could follow the console repo's approach: clone the cloud repo in CI and start the local development environment, which provides a testing Frontegg environment, a local cloud API, and a local Materialize environment.
  * You could always run CI against staging.

I'm honestly not sure which is best.

benesch (commented):
Looking good! Dropped in a few comments.

docs/developer/design-v2-2023-11-11.md
bobbyiliev: As far as I can tell the Terraform's official documentation doesn't explicitly detail best practices for hiding provider configuration options. Probably because typically, all configuration options are intended to be exposed to users.

Maybe an option here is to remove the endpoint from the provider configuration schema, and only expose it as an `MZ_ENDPOINT` env variable. We can document that in the dev docs. Open to suggestions @dehume

benesch: Ah yeah I was imagining it would work like the `sslmode` option today, where it's available as a configuration option but we just hide it in the docs. Only supporting it via the `MZ_ENDPOINT` environment variable sounds like a reasonable option too.

dehume: What Nikhil said. We entirely override the [main provider page](https://github.com/MaterializeInc/terraform-provider-materialize/blob/main/docs/index.md). So it would not be immediately exposed to users unless we add it.

bobbyiliev: Sounds good! I will keep it as the `sslmode` option in this case!

docs/developer/design-v2-2023-11-11.md
benesch: Yeah, the mocks you added in this PR look good!

benesch (commented):
Nice! I only took a very quick skim, but this all seems to be headed in a wonderful direction. 👍🏽

pkg/resources/resource_user.go
```diff
@@ -0,0 +1,298 @@
...
+
+	// Decode the JSON response
+	var rolesResponse FronteggRolesResponse
+	if err := json.NewDecoder(resp.Body).Decode(&rolesResponse); err != nil {
+		return nil, fmt.Errorf("error decoding response: %v", err)
+	}
+
+	// Create a map of role names to their IDs
+	roleMap := make(map[string]string)
+	for _, role := range rolesResponse.Items {
```
dehume, line 288: Does this work with multiple roles? Won't the map just override the latest `Admin` and `Member` keys with the latest role id with those roles?

dehume (commented):
🚀 This is amazing. A lot went into it. I left some comments focused most on the new resources/data sources. Still digesting a lot of it. Will make another pass later.

pkg/resources/resource_user.go
```diff
@@ -0,0 +1,298 @@
...
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+			"roles": {
+				Type:        schema.TypeList,
+				Elem:        &schema.Schema{Type: schema.TypeString},
+				Required:    true,
+				ForceNew:    true,
+				Description: "The roles to assign to the user. Allowed values are 'Member' and 'Admin'.",
```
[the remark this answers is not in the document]
bobbyiliev: Yep, I initially had that but I hit an issue where the│ `ValidateFunc` and `ValidateDiagFunc` are not yet supported on lists or sets :/

image

Hance that messy workaround with the description message and the check within the `userCreate` func.

pkg/resources/resource_user.go
bobbyiliev: The idea behind this is to map the shorthand syntax eg. `Admin` and `Member` to the actual response from the API, eg:

```
Roles response: {[{11111111-aaaaa-11111-aaaaa-1111111 Organization Admin} {22222222-bbbbb-11111-aaaa-2222222 Organization Member}] {2 1}}
```

That way in the schema we could require `Admin` and/or `Member` instead of asking for the `Organization` prefix as well.

Probably there is a better way of handing this!

pkg/resources/resource_app_password.go
```diff
@@ -0,0 +1,280 @@
...
+	if err := json.NewDecoder(resp.Body).Decode(&response); err != nil {
+		return response, err
+	}
+
+	return response, nil
+}
+
+// Helper function to construct the full API endpoint for app passwords
+func getApiEndpoint(client *clients.FronteggClient, resourcePath string, resourceID ...string) string {
+	if len(resourceID) > 0 {
```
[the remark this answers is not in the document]
bobbyiliev: The idea here is to be able to either provide 1 `resourceID` or 0. That way the function can be used with either 2 or 3 parameters when needed, eg:

```
getApiEndpoint(client, apiTokenPath, resourceID)
getApiEndpoint(client, apiTokenPath)
```

But as Go does not have optional parameters for functions, that was my workaround. Totally open to suggestions!

dehume (approved):
This looks good. Should be an exciting `0.5.0` release. The last thing I would update is the [template index ](https://github.com/MaterializeInc/terraform-provider-materialize/blob/main/templates/index.md.tmpl) to reflect the new provider attribute names

pkg/clients/frontegg_client.go
```diff
@@ -0,0 +1,238 @@
...
+
+// Get the token expiry from the FronteggClient
+func (c *FronteggClient) GetTokenExpiry() (time.Time, error) {
+	return c.TokenExpiry, nil
+}
+
+// Get the password from the FronteggClient
+func (c *FronteggClient) GetPassword() (string, error) {
+	return c.Password, nil
+}
```
sjwiesman, line 168: Why do these return an error if it's always nil?

pkg/clients/frontegg_client.go
```diff
@@ -0,0 +1,238 @@
...
+	}
+	return strings.Join(parts, "-")
+}
+
+func (c *FronteggClient) NeedsTokenRefresh() bool {
+	return time.Now().After(c.TokenExpiry)
+}
+
+func (c *FronteggClient) RefreshToken() error {
+	log.Printf("[DEBUG] Refreshing Frontegg: %v\n", c)
```
sjwiesman, line 218: Did you intend to keep this?

pkg/clients/frontegg_client.go
```diff
@@ -0,0 +1,238 @@
...
+type tokenTransport struct {
+	Token     string
+	Transport http.RoundTripper
+}
+
+// RoundTrip method to execute the request with the token
+func (t *tokenTransport) RoundTrip(req *http.Request) (*http.Response, error) {
+	// Deep copy the request to ensure it's safe to modify
+	req2 := cloneRequest(req)
+	req2.Header.Set("Authorization", "Bearer "+t.Token)
```
sjwiesman, line 64: Can you say more about why we need the deep copy?

pkg/utils/provider_meta.go
```diff
@@ -0,0 +1,94 @@
...
+	DefaultRegion  clients.Region
+	RegionsEnabled map[clients.Region]bool
+}
+
+var DefaultRegion string
+
+func GetProviderMeta(meta interface{}) (*ProviderMeta, bool) {
+	providerMeta, ok := meta.(*ProviderMeta)
+	if !ok || providerMeta == nil {
+		fmt.Println("Type assertion failed: provider meta is not of type *ProviderMeta or is nil")
```
sjwiesman, line 25: Should this be a panic?

pkg/utils/provider_meta.go
```diff
@@ -0,0 +1,94 @@
...
+		return nil, false
+	}
+	if providerMeta.Frontegg.NeedsTokenRefresh() {
+		err := providerMeta.Frontegg.RefreshToken()
+		if err != nil {
+			fmt.Printf("Failed to refresh token: %v\n", err)
```
sjwiesman, line 31: Delete?

pkg/utils/provider_meta.go
```diff
@@ -0,0 +1,94 @@
...
+import (
+	"fmt"
+	"strings"
+
+	"github.com/MaterializeInc/terraform-provider-materialize/pkg/clients"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/jmoiron/sqlx"
+)
+
+type ProviderMeta struct {
```
sjwiesman, line 12: This struct seems important. Maybe add some additional comments?

pkg/utils/provider_meta.go
benesch: Agreed; seems like this could just be `providerMetda := meta.(*ProviderMeta)`.

pkg/utils/provider_meta.go
benesch: Why does this function return a `bool` instead of an `error`? Seems like this just wants to return the error in this case, and then the caller can wrap the error with more context.

pkg/clients/frontegg_client.go
benesch: (This is just how you emit debug logs in Terraform Providers: https://developer.hashicorp.com/terraform/plugin/log/writing#debug-1)

pkg/clients/frontegg_client.go
bobbyiliev: This is primarily related to concurrency and the immutability of the original request. As http.Request.Body can only be read once, a new body needs to be copied:

> https://stackoverflow.com/questions/62017146/http-request-clone-is-not-deep-clone

Will try to adjust the comment to give a bit more details!

sjwiesman: How about you include the stack overflow link in the comment.

pkg/utils/provider_meta.go
bobbyiliev: Good point! Just added some comments.

pkg/utils/provider_meta.go
bobbyiliev: Good catch! Just changed this to actually return an error so we could use the standard:

```golang
if err != nil {
	return diag.FromErr(err)
}
```

pkg/utils/provider_meta.go
bobbyiliev: Just updated this accordingly, now you can use this similar to other functions in the resources and data sources, eg:

```golang
providerMeta, err := utils.GetProviderMeta(meta)
if err != nil {
	return diag.FromErr(err)
}
```

pkg/clients/frontegg_client.go
bobbyiliev: Sounds good, just added the link in the func comments.

pkg/clients/frontegg_client.go
bobbyiliev: Ah yes, seems like I've missed this after one of the reactors! Good catch.

PR #353, opened 2023-11-09
Prevent Attribute Name Leaking

### LINKED ISSUE #352: Leaking `database_name` and `schema_name` in Resources
Resources will leak their top level `database_name` and `schema_name` attributes into other reference blocks. For example if a connection relies on a password stored in another database:

```
resource "materialize_connection_postgres" "postgres_connection" {
  name = "my_connection"
  schema_name  = "my_schema"
  database_name  = "my_database"

  password {
    name = "my_password"
    database_name = "another_database"
  }
}
```

The generated DDL will infer the schema set for the connection resource and attempt to use `"another_database"."my_schema"."my_password"` as the password which may not exist and can be confusing for the user.

dehume:
Fixes #352

Prevent leaking of top level `database_name` and `schema_name` resource attributes into other references that have a database and schema names. Instead of inheriting these (if not set), they will use the standard defaults for schema (`public`) and database (`materialize` or the env variable).


pkg/materialize/identifiers.go
```diff
@@ -14,15 +14,11 @@ func GetIdentifierSchemaStruct(databaseName string, schemaName string, v interfa
 	if v, ok := u["name"]; ok {
```
benesch, line 14: Should this function even take `databaseName` or `schemaName` as input now? They look unused now.

benesch (commented):
This fix makes a lot of sense to me. This may be enough of a breaking change that it needs to be a 0.3.0 though!

Not approving as I didn't look in detail at the code.

bobbyiliev (approved):
Looks good! Thank you for working on this and for introducing the fix so quickly!

PR #361, opened 2023-11-14
Add NOT ENFORCED for sink keys

bobbyiliev:
Adds a missing `NOT ENFORCED` option to the sink resource.

> https://materialize.com/docs/sql/create-sink/kafka/


docs/resources/sink_kafka.md
```diff
@@ -64,6 +64,7 @@ resource "materialize_sink_kafka" "example_sink_kafka" {
 - `envelope` (Block List, Max: 1) How to interpret records (e.g. Debezium, Upsert). (see [below for nested schema](#nestedblock--envelope))
 - `format` (Block List, Max: 1) How to decode raw bytes from different formats into data structures it can understand at runtime. (see [below for nested schema](#nestedblock--format))
 - `key` (List of String) An optional list of columns to use for the Kafka key. If unspecified, the Kafka key is left unset.
+- `not_enforced` (Boolean) Disable Materialize's validation of the key's uniqueness.
```
benesch, line 67: I think you want to call this `key_not_enforced` to make it clear that it's the key that's not enforced. 

I think actually ideally this would have been a nested schema...

```
key {
    columns = ["a", "b"]
    not_enforced = true
}
```

...but that's a non backwards incompatible change that's probably not worth it.

dehume (approved):
Looks good. Interesting the docs Github Action didn't run again after your last commit. The docs still show `not_enforced` rather than `key_not_enforced`

benesch (approved):

PR #363, opened 2023-11-15
Add ForceNew for table column name changes

bobbyiliev:
As discussed [here](https://materializeinc.slack.com/archives/C015RHB3LDR/p1699958008429369), adding a `ForceNew` for table column name changes as we do not yet support `RENAME COLUMN`:

> https://github.com/MaterializeInc/materialize/issues/5301

### DIFF
--- a/pkg/provider/acceptance_table_test.go
+++ b/pkg/provider/acceptance_table_test.go
@@ -76,6 +76,30 @@ func TestAccTable_update(t *testing.T) {
 					resource.TestCheckResourceAttr("materialize_table.test_role", "ownership_role", roleName),
 				),
 			},
+			{
+				Config: testAccTableResourceWithUpdatedColumnName(roleName, tableName, tableRoleName, "mz_system", "new_column_1"),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckTableExists("materialize_table.test"),
+					resource.TestCheckResourceAttr("materialize_table.test", "name", tableName),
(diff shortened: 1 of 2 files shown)


dehume (changes requested):
Thanks for doing this. While we are doing this I think we need to make two other changes:

* We should include a `ForceNew` for both `type` and `nullable` as well. The nested `ForceNew` is confusing but right now I think it would trigger a force new on new columns being added OR changing the name of an existing column. So we still want to catch changes on an existing column if it is type or nullable
* The only column update we support are comments, so comments do not need a `ForceNew` but the update it [here](https://github.com/MaterializeInc/terraform-provider-materialize/blob/main/pkg/resources/resource_table.go#L235) to `column`. Since currently it will never trigger.

bobbyiliev:
> Thanks for doing this. While we are doing this I think we need to make two other changes:
> 
> * We should include a `ForceNew` for both `type` and `nullable` as well. The nested `ForceNew` is confusing but right now I think it would trigger a force new on new columns being added OR changing the name of an existing column. So we still want to catch changes on an existing column if it is type or nullable
> * The only column update we support are comments, so comments do not need a `ForceNew` but the update it [here](https://github.com/MaterializeInc/terraform-provider-materialize/blob/main/pkg/resources/resource_table.go#L235) to `column`. Since currently it will never trigger.

Ah yes! very good catch, I missed this initially as I only was testing the table comments rather than column comments. I've made some changes to this now and it seems to be working as expected. Let me know what you think @dehume!

dehume (approved):

PR #364, opened 2023-11-15
Update Webhook Example

dehume:
Update the integration and docs examples for the source webhook. It can be particularly confusing around how to set `check_options`.


bobbyiliev (approved):
Looks good! Thank you for updating this!

PR #370, opened 2023-11-16
Replica Deprecation Warning

dehume:
It is unclear that the cluster replica resource should not be used instead of managed clusters.  Adding a similar deprecation warning to the `materialize_cluster_replica` resource that exists in the [docs](https://materialize.com/docs/sql/create-cluster-replica/).

Also looking to make the `materialize_cluster` resource more clear around creating managed clusters.


bobbyiliev (approved):
Very nice!

dehume:
Merging. Will put out a separate PR around the flakiness of the acceptance tests.

PR #371, opened 2023-11-17
Raise Max Clusters Testing

### LINKED ISSUE #347: Update `max_clusters` for Acceptance Tests
The acceptance tests can fail occasionally because the number of active clusters is reached (10). For integration and acceptance tests the `max_clusters` should be set to something higher to avoid this.

dehume:
Fixes #347

Raises the cap on the number of clusters allowed for the acceptance tests

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -19,6 +19,18 @@ services:
       - 6878:6878
     healthcheck: {test: curl -f localhost:6878/api/readyz, interval: 1s, start_period: 35s}
 
+  init:
+    image: postgres
+    container_name: materialized_init
+    depends_on:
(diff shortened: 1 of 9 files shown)


bobbyiliev (commented):
Seems like the tests are failing. If you don’t come up with a work around, here is how I did it with the metabase tests:

https://github.com/MaterializeInc/metabase-materialize-driver/blob/master/docker-compose.yml#L31

dehume:
> If you don’t come up with a work around, here is how I did it with the metabase tests:

That is much cleaner. Implementing it that way. Thanks

bobbyiliev (approved):

PR #372, opened 2023-11-20
Handle usernames with special characters in them

benesch:
Which, in practice, is all usernames in production, since they're all email addresses that contain an `@` sign.

Also:

  * Make the `sslmode` parameter a string rather than a Boolean. It's a hidden parameter for testing, so might as well allow the full power of the underlying `sslmode` option. Makes the code simpler.

  * Rename `application_name` to `application_name_suffix`, and force include "terraform-provider-materialize vX.Y.Z" in the name. While this is also a hidden parameter, no point allowing someone to easily mess with our metrics by changing the application name. The `application_name_suffix` parameter allows the downstream Pulumi provider to inject "(Pulumi version ...)" into the application name. Possibly it could also be useful to end users, if they want to distinguish between different Terraform projects.

----

@dehume when this merges, a new release of the Pulumi provider would be great. We're bumping into this as we set up mz-on-mz: https://github.com/MaterializeInc/i2/pull/1370/files#diff-24af2d7896d8dfce91831fe5a92050e971abd23b8ca9192d4bd17b2a852b4f67R51-R54

cc @necaris

### DIFF
--- a/.github/workflows/acceptance.yml
+++ b/.github/workflows/acceptance.yml
@@ -22,7 +22,7 @@ jobs:
         env:
           MZ_HOST: localhost
           MZ_USER: mz_system
-          MZ_SSLMODE: "false"
+          MZ_SSLMODE: disable
           MZ_PORT: 6877
 
       - name: Docker Compose Down
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Changelog
 
+## Unreleased
+
+### BugFixes
+
+* Handle `user` values that contain special characters, without requiring manual
+  URL escaping (e.g., escaping `you@corp.com` as `you%40corp.com`).
+
 ## 0.3.0 - 2023-11-16
 
 ### Features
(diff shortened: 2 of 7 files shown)


dehume (approved):
> The `application_name_suffix` parameter allows the downstream Pulumi provider to inject "(Pulumi version ...)" into the application name. Possibly it could also be useful to end users, if they want to distinguish between different Terraform projects.

Ok. I was handling `application_name` that way since I thought the only way to inject the pulumi name was overriding the provider level configs: https://github.com/MaterializeInc/pulumi-materialize/blob/main/provider/resources.go#L67-L70

bobbyiliev (approved):

benesch:
That is the only way that I'm aware of! I just wanted to avoid letting users entirely blow away the `terraform-provider-materialize` part of the application name. The Pulumi provider can then set an `application_name_suffix` of `(pulumi-materialize v...)`, and the full application name will read `terraform-provider-materialize v0.1.0 (pulumi-materialize v0.1.0)`.

PR #373, opened 2023-11-20
Support Avro Doc Comments `materialize_sink_kafka`

### LINKED ISSUE #359: Kafka sink: support `DOC ON` options
[Materialize #22019](https://github.com/MaterializeInc/materialize/pull/22019) introduced support for adding comments to the Avro schema generated with `CREATE SINK` using the following syntax:

```sql
CREATE TABLE t (c1 int, c2 text);
COMMENT ON TABLE t IS 'materialize comment on t';
COMMENT ON COLUMN t.c2 IS 'materialize comment on t.c2';

CREATE SINK avro_sink
  IN CLUSTER my_io_cluster
  FROM t
  INTO KAFKA CONNECTION kafka_connection (TOPIC 'test_avro_topic')
  KEY (c1)
  FORMAT AVRO USING CONFLUENT SCHEMA REGISTRY CONNECTION csr_connection
  (
    DOC ON TYPE t = 'top-level comment for avro record in both key and value schemas',
    KEY DOC ON COLUMN t.c1 = 'comment on column only in key schema',
    VALUE DOC ON COLUMN t.c1 = 'comment on column only in value schema'
  )
  ENVELOPE UPSERT;
```

See the [feature documentation](https://materialize.com/docs/sql/create-sink/kafka/#avro-comments) for more details.

dehume:
Fixes #359

Add two new attributes to the `materialize_sink_kafka`, `avro_doc_type` and `avro_doc_column` to support [Avro comments](https://materialize.com/docs/sql/create-sink/kafka/#avro-comments). Adds the following syntax to the SQL
```
CREATE SINK avro_sink
  ...
  (
    DOC ON TYPE t = 'top-level comment for avro record in both key and value schemas',
    KEY DOC ON COLUMN t.c1 = 'comment on column only in key schema',
    VALUE DOC ON COLUMN t.c1 = 'comment on column only in value schema'
  )
```


benesch:
This is still WIP, so let me know if this is too deep too soon, but I think we should track the SQL syntax more closely here. I think you want something more like:

```hcl
format {
    avro {
        doc_on_type {
            database_name = "..."
            schema_name = "..."
            type_name = "..."
            doc = "..."
            # At least one of key or value must be true.
            key = {true | false}
            value = {true | false}
        }

        doc_on_column {
            database_name = "..."
            schema_name = "..."
            item_name = "..."
            column_name = "..."
            doc = "..."
            # At least one of key or value must be true.
            key = {true | false}
            value = {true | false}
        }
    }
}
```

pkg/materialize/sink_kafka.go
```diff
@@ -130,8 +130,45 @@ func (b *SinkKafkaBuilder) Create() error {
...
+		// Doc Type
+		if b.format.Avro.DocType.Object.Name != "" {
+			c := strings.Builder{}
+			if b.format.Avro.DocType.Key {
+				c.WriteString("KEY ")
+			} else if b.format.Avro.DocType.Value {
+				c.WriteString("VALUE ")
+			}
+			c.WriteString(fmt.Sprintf("DOC ON TYPE %[1]s = %[2]s",
+				b.format.Avro.DocType.Object.QualifiedName(),
```
dehume, line 152: Based on docs, unsure if we need to also support aliases for the object reference. Or if it should just always make the connection.

pkg/resources/schema.go
```diff
@@ -285,6 +285,66 @@ func SinkFormatSpecSchema(elem string, description string, required bool) *schem
 								Optional:    true,
 								ForceNew:    true,
 							},
+							"avro_doc_type": {
+								Description: "**Private Preview** Add top level documentation comment to the generated Avro schemas.",
```
dehume, line 289: Open to better descriptions for all these.

dehume:
The acceptance test failing is within Kafka Sources, not sinks. Must be due to some of the shared  structs within `format_specs`

pkg/materialize/format_specs.go
```diff
@@ -89,10 +106,42 @@ func GetSinkFormatSpecStruc(v interface{}) SinkFormatSpecStruct {
 		if csr, ok := avro.([]interface{})[0].(map[string]interface{})["schema_registry_connection"]; ok {
 			key := avro.([]interface{})[0].(map[string]interface{})["avro_key_fullname"].(string)
 			value := avro.([]interface{})[0].(map[string]interface{})["avro_value_fullname"].(string)
+
+			var docType AvroDocType
+			if adt, ok := avro.([]interface{})[0].(map[string]interface{})["avro_doc_type"].([]interface{}); ok {
```
bobbyiliev, line 111: I think that the tests are failing due to this line here. As far as I can tell `adt[0]` is accessed without checking if adt (which is expected to be a slice) has any elements. 

```suggestion
			if adt, ok := avro.([]interface{})[0].(map[string]interface{})["avro_doc_type"].([]interface{}); ok && len(adt) > 0 {
```

bobbyiliev (approved):

PR #378, opened 2023-11-21
Prevent Privilege Read Errors

dehume:
Cleaning up the grant code which can lead to cryptic errors. Part of this was handling the privileges column from the system catalog as a string (now using `pq.StringArray`). Issues can come from attempting to generate the compound key that makes up grants and parsing out the compound key in order to correctly query the mz catalog.
```
panic: runtime error: index out of range [0] with length 0
```
If any issues arise with the parsing the key, the provider will set the key to `""` so the grant can be recreated rather than erroring.

Make the grant parsing functions more consistent and simple:
* `PrivilegeName`
* `ParseMzAclString`
* `MapGrantPrivileges`
* `MapDefaultGrantPrivileges`

### DIFF
--- a/.github/workflows/acceptance.yml
+++ b/.github/workflows/acceptance.yml
@@ -14,6 +14,13 @@ jobs:
     steps:
       - uses: actions/checkout@v3
 
+      - uses: actions/setup-go@v4
+        with:
+          go-version-file: "go.mod"
+          cache: true
+
+      - run: go mod download
+
       - name: Docker Compose Up
         run: docker compose up --build -d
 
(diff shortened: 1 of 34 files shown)


dehume:
Have to look into why the GHAs aren't picking up the package change

dehume:
Any suggestion on the new `BrokerTransportFailure` failures. I didn't change anything around kafka connections

bobbyiliev (changes requested):
As discussed internally, it sounds like that we might have to enable `security_protocol = "SASL_PLAINTEXT"` in the `connection.tf` to fix the failing tests

bobbyiliev (approved):
Looks good!

PR #380, opened 2023-11-22
Set Security Protocol

dehume:
Now that `security protocol` is enabled, setting so the integration and acceptance tests will pass. Also updating the flakey table grant acceptance test.


bobbyiliev (approved):
Neat! Thank you!

PR #381, opened 2023-11-22
Default Grant Read Unnecessary `Set`

dehume:
Remove unnecessary `d.Set` from `grantDefaultPrivilegeRead`. Reads for grants never need to set any attributes, it just needs to confirm that the query can be preformed based on the compound key and that the privilege is present in the returned result. Removing this part prevents an error in the provider if the query returns no rows. 

Updating the RBAC integration tests to be a more complex matrix to more throughly test.

### DIFF
--- a/integration/rbac.tf
+++ b/integration/rbac.tf
@@ -49,7 +49,7 @@ resource "materialize_role" "target_2" {
 variable "database_grants" {
   type = map(any)
   default = {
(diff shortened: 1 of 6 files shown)


bobbyiliev (approved):
I've been testing this locally and so far it seems to be working as expected. Nice job!

Issue #368, opened 2023-11-16
[EPIC] Allow `ALTER CONNECTION` Updates


dehume:
[Materialize #21109](https://github.com/MaterializeInc/materialize/pull/21109) will allow `ALTER CONNECTION` for all available options for each connector. All connection resources will need to update their `UpdateContext` to allow for all of the possible updates that can occur.
```[tasklist]
### Tasks
- [ ] https://github.com/MaterializeInc/terraform-provider-materialize/issues/513
- [ ] https://github.com/MaterializeInc/terraform-provider-materialize/issues/514
- [ ] https://github.com/MaterializeInc/terraform-provider-materialize/issues/515
- [ ] https://github.com/MaterializeInc/terraform-provider-materialize/issues/516
- [ ] https://github.com/MaterializeInc/terraform-provider-materialize/issues/517
- [ ] https://github.com/MaterializeInc/terraform-provider-materialize/issues/518
- [ ] https://github.com/MaterializeInc/terraform-provider-materialize/issues/519
```

bobbyiliev:
Closing this one as completed

PR #388, opened 2023-11-27
Expand Connections Test Coverage (Connections)

bobbyiliev:


dehume (approved):
Looking good. The additional tests in the `integration/connection.tf` should be helpful. The thing that will be hardest with connections is the recent work with `ALTER CONNECTION...` (https://github.com/MaterializeInc/terraform-provider-materialize/issues/368). But we can keep that to a separate issue and include all the additional acceptance tests required then.

bobbyiliev:
Sounds good! Will grab that issue now.

PR #392, opened 2023-11-30
Update Changelog for v0.3.3

bobbyiliev:


CHANGELOG.md
```diff
@@ -2,6 +2,14 @@
 
 ## Unreleased
 
+### Features
```
morsapaes, line 5: It's a bit weird to include the blank sections for an unreleased version! I'd add these whenever there is a commit that add requires a changelog entry after the release is cut, not when cutting the release (that's what we do for dbt).

dehume (approved):

morsapaes (approved):

PR #396, opened 2023-12-04
PoC for IDs migration for the database resource

bobbyiliev:


pkg/utils/ids_migration_helper.go
```diff
@@ -0,0 +1,62 @@
...
+	}
+	// If the ID already has a region, return the original ID
+	if strings.Contains(oldID, ":") {
+		return oldID, nil
+	}
+	return fmt.Sprintf("%s:%s", region, oldID), nil
+}
+
+// Function to get the ID from the region + ID string
+func ExtractId(oldID string) (string, error) {
```
dehume, line 40: Seems like the error return is `nil` in both cases, maybe remove.

pkg/resources/resource_database.go
```diff
@@ -101,6 +122,12 @@ func databaseCreate(ctx context.Context, d *schema.ResourceData, meta interface{
 	if err != nil {
 		return diag.FromErr(err)
 	}
+
+	// set id with the region
```
dehume, line 126: So we have to set `TransformIdWithRegion` in both the create and read for state imports and the initial migration?

pkg/provider/provider.go
```diff
@@ -131,6 +132,9 @@ func providerConfigure(ctx context.Context, d *schema.ResourceData, version stri
 	sslmode := d.Get("sslmode").(string)
 	application_name := fmt.Sprintf("terraform-provider-materialize v%s", version)
 
+	// Set the host in the utils package so that the region can be extracted from it
+	utils.Host = host
```
dehume, line 136: This solution is just temporary for the migration. host is would be removed from the provider entirely when cloud resources are added?

pkg/materialize/database.go
```diff
@@ -71,6 +72,9 @@ func DatabaseId(conn *sqlx.DB, obj MaterializeObject) (string, error) {
 }
 
 func ScanDatabase(conn *sqlx.DB, id string) (DatabaseParams, error) {
+	// Get the database id from the state id
+	id, _ = utils.ExtractId(id)
```
dehume, line 76: Nit: But I would remove the `ExtractId` from within the Scan functions. I feel like the `materialize` package should try to be separate from the Terraform functionality and combining them makes it harder to follow. We may also want to think of naming our variables around `id` since there is now the id within Materialize and the Terraform id.

dehume (commented):
Overall I think this makes sense. I realize this is slightly a bandaid until we get cloud resources in so I might be slightly overcautious but trying to still keep the packages separate as much as possible.

pkg/utils/ids_migration_helper.go
```diff
@@ -0,0 +1,62 @@
+package utils
+
+import (
+	"context"
+	"fmt"
+	"strings"
+)
+
+var Host string
```
benesch, line 9: I'd recommend instead storing `var Region string` here, and when you set the parameter during initialization you can extract the region once then, rather than extracting it from the hostname every time `TransfromIdFromRegion` is called. Also then you can error during initialization instead of in `TransformIdWithRegion`, so you won't need to return an error there.

pkg/materialize/database.go
benesch: Agreed. Better to have the `id` here be a Materialize database ID, IMO, rather than a TF provider database ID.

pkg/utils/ids_migration_helper.go
```diff
@@ -0,0 +1,62 @@
...
+	region := ExtractRegionFromHostname()
+	if region == "" {
+		return "", fmt.Errorf("failed to extract region from hostname")
+	}
+	// If the ID already has a region, return the original ID
+	if strings.Contains(oldID, ":") {
+		return oldID, nil
+	}
+	return fmt.Sprintf("%s:%s", region, oldID), nil
+}
```
benesch, line 37:
```suggestion
func TransformIdWithRegion(oldID string) string {
	// If the ID already has a region, return the original ID
	if strings.Contains(oldID, ":") {
		return oldID
	}
	return fmt.Sprintf("%s:%s", Region, oldID)
}
```

pkg/utils/ids_migration_helper.go
```diff
@@ -0,0 +1,62 @@
...
+	}
+
+	newID, err := TransformIdWithRegion(oldID)
+	if err != nil {
+		return nil, err
+	}
+	rawState["id"] = newID
+
+	return rawState, nil
+}
```
benesch, line 62: This function LGTM.

pkg/resources/resource_database.go
```diff
@@ -46,6 +62,11 @@ func databaseRead(ctx context.Context, d *schema.ResourceData, meta interface{})
 		return diag.FromErr(err)
 	}
 
+	// set id with the region
+	i, err = utils.TransformIdWithRegion(i)
+	if err != nil {
+		return diag.FromErr(err)
+	}
 	d.SetId(i)
```
benesch, line 70: (see below)

```suggestion
	d.SetId(utils.TransformIdWithRegion(i))
```

benesch (commented):
Nice! This looks pretty clean.

pkg/provider/provider.go
bobbyiliev: Yep exactly. The idea is to not have to ask the user to specify it manually.

pkg/resources/resource_database.go
bobbyiliev: Yes, I _think_ that this is the case. We should use `TransformIdWithRegion` in both the create and read functions as it ensures consistency in how IDs are managed and stored in the Terraform state for new resources.

pkg/materialize/database.go
bobbyiliev: @benesch @dehume just made all of the requested changes. Let me know what you think. If this is ok, I will start with all of the rest of the resources.

benesch: > just made all of the requested changes. Let me know what you think.

Looks great to me!

> If this is ok, I will start with all of the rest of the resources.

Godspeed.

dehume (commented):
One last thing I thought of. We should probably update all the [example import.sh](https://github.com/MaterializeInc/terraform-provider-materialize/blob/main/examples/resources/materialize_database/import.sh) to include region. The good thing is this will remain true after the cloud resources get merged (also was able to successfully run an import using the region)
```
resource "materialize_database" "import" {
  name    = "import"
}
```
```
terraform import materialize_database.import aws/us-east-1:u5
```

dehume (approved):

PR #397, opened 2023-12-05
Allow `PUBLIC` for `grantee` with Default Grants

dehume:
The documentation indicates that `PUBLIC` can be used for both `target_role` and `grantee` for default grant resources. However this was only true for `target_role`. Supporting `grantee` as well.


bobbyiliev (approved):
Thank you for the quick fix!

PR #406, opened 2023-12-11
Grant Id Set Read

dehume:
The grant read functions was not setting the new id value which would mean that existing grants would not be properly updated after the migration to `0.4.0`. Including `TransformIdWithRegion` in the `SetId` on all of the read functions. Also while including the unit tests for the grant read functions, including a similar migration id unit test for all other resources.


bobbyiliev (approved):
Just tested this locally, and it seems to be all working as expected now!

PR #410, opened 2023-12-12
Duplicate Datasources

dehume:
Investigation for duplicate datasources. It seems that the issue was with `materialize_materialized_views` that contained a top level comment as well as column level comments. This cannot be set via the provider (mostly likely dbt) which was leading to duplicated rows in the data sources. Including `AND object_sub_id IS NULL` to the subquery removes the column level comments.

Also including acceptance tests for all data sources:
| Datasource | Status |
| --- | --- |
| cluster replica | X |
| cluster | X |
| connection | X |
| current cluster | X |
| current database | X |
| database | X |
| egress ip | X |
| index | X |
| materialized view | X |
| role | X |
| schema | X |
| secret | X |
| sink | X |
| source | X |
| table | X |
| type | X |
| view | X |


pkg/utils/ids_migration_helper_test.go
```diff
@@ -26,7 +26,7 @@ func TestTransformIdWithRegion(t *testing.T) {
 			"expected": "aws/us-east-1:u1",
 		},
 	}
-	for tc, _ := range testCases {
+	for tc := range testCases {
```
dehume, line 29: Unrelated, just from running format

bobbyiliev (changes requested):
Seems like that the IDs migration CI is failing, but it should no longer be needed as the version has now been released.

CHANGELOG.md
```diff
@@ -2,6 +2,12 @@
 
 ## Unreleased
 
+### Features
+* Allow Avro comments (`avro_doc_type` and `avro_doc_column`) for resource `materialize_sink_kafka` [#373](https://github.com/MaterializeInc/terraform-provider-materialize/pull/373)
```
dehume, line 6: Was not included in the last PR

pkg/materialize/materialized_view.go
```diff
@@ -114,6 +114,7 @@ var materializedViewQuery = NewBaseQuery(`
 		SELECT id, comment
 		FROM mz_internal.mz_comments
 		WHERE object_type = 'materialized-view'
+		AND object_sub_id IS NULL
```
dehume, line 117: This is the actual fix for the duplication issue. The majority of the PR is just testing

bobbyiliev (approved):
This is great! Thank you for adding all of those tests!

PR #418, opened 2023-12-18
Kafka Acc Test Offset

dehume:
The Kafka source in the image now seems to validate that the offset matches with the number of partitions. Changing in the acceptance tests so those tests will pass.

Also preventing the destroy for table default grants which is flaky when running snippets of grants. Destroy is more reliably captured in the integration testing.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -19,7 +19,7 @@ services:
       - 6878:6878
     healthcheck: {test: curl -f localhost:6878/api/readyz, interval: 1s, start_period: 35s}
 
-  init:
+  materialized_init:
     image: postgres
     container_name: materialized_init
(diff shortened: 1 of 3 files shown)


bobbyiliev (approved):
Thank you for sorting that out 🙏

morsapaes:
This change was introduced in #23506.

dehume:
The issue now is that the topic needs to exist as well. It seems like this isn't a problem with the integration tests, probably some combination of resources in the project is creating the topic. Looking for the best way to handle this for the acceptance tests.

dehume:
Can look into a better solution later but fixing for now by having the acceptance test for the Kafka source create a topic that it uses for its tests.

Issue #421, opened 2023-12-19
Failed PostgreSQL Source update causes invalid state and won't refresh


matthelm:
We were recently adding new sub-sources to an existing PostgreSQL source and had a bumpy experience as follows:
- The deploy adding the sub-sources failed with the following error: `ERROR: table raw.organization__c not found in source (SQLSTATE XX000)`. We realized we had a typo in our sub-source names.
- We deployed a revert, removing the new sub-sources, but were surprised to see that it also failed. `ERROR: table raw.organization__c not found in source (SQLSTATE XX000)`.
- Looking at the Materialize instance we confirmed that the sub-sources hadn't been created.
- Looking at Pulumi's state we found that the sub-sources were listed, which seemed incorrect.
- We tried a `pulumi refresh` but it surprisingly reported there was no changes.
- We fixed the issue by manually editing the state, removing the sub-sources and the error message.

This leaves us suspecting two potential bugs with the PostgreSQL source resource type:
1. A failed attempt to add sub-sources still adds them to the state
2. A `pulumi refresh` doesn't reload the sub-sources

benesch:
Something seems funny with the way the Terraform provider is rigged up on this. We provide the `tables` parameter as input...

https://github.com/MaterializeInc/i2/blob/bdfb53146ea264957bbb13d868041fdaa3d5e885/i2/odw/resource.py#L677

but it's actually this `subsource` parameter that is updated on read:

https://github.com/MaterializeInc/terraform-provider-materialize/blob/6b3637a64db28c90d1766e4b3c35172cf68ece84/pkg/resources/resource_source_postgres.go#L67C31-L67C31

https://github.com/MaterializeInc/terraform-provider-materialize/blob/6b3637a64db28c90d1766e4b3c35172cf68ece84/pkg/resources/resource_source.go#L61-L80

That `subsource` parameter is totally ignored by `sourcePostgresUpdate`, which explains why the `pulumi refresh` had no effect. We can limit our investigation to this code in `sourcePostgresUpdate`, which handles addition/removal of subsources based on changes to the `tables` parameter:

https://github.com/MaterializeInc/terraform-provider-materialize/blob/6b3637a64db28c90d1766e4b3c35172cf68ece84/pkg/resources/resource_source_postgres.go#L203-L224

[...]

PR #422, opened 2023-12-20
[WIP] Postgres Subsource Updates

dehume:
Fixes https://github.com/MaterializeInc/terraform-provider-materialize/issues/421

1. Include `table` as part of the read for the resource to prevent issues when an improper table is attempted to be added.
2. Remove the options for `FOR SCHEMAS (...)` and `FOR ALL TABLES` and require the user to explicitly set every table they would like synced with the source.


pkg/resources/resource_source_postgres.go
```diff
@@ -34,7 +35,7 @@ var sourcePostgresSchema = map[string]*schema.Schema{
 		Optional:    true,
 	},
 	"table": {
-		Description: "Creates subsources for specific tables. If neither table or schema is specified, will default to ALL TABLES",
+		Description: "Creates subsources for specific tables in the Postgres connection.",
 		Type:        schema.TypeList,
```
dehume, line 39: This will need to be converted to `TypeSet` to ignore order which cannot be guaranteed between how the user enters the `table` values and what the system catalog query will return.

pkg/resources/resource_source_postgres.go
```diff
@@ -85,6 +77,86 @@ func SourcePostgres() *schema.Resource {
...
+	deps, err := materialize.ListDependencies(meta.(*sqlx.DB), utils.ExtractId(i), "source")
+	if err != nil {
+		return diag.FromErr(err)
+	}
+
+	// Tables
+	tMaps := []map[string]interface{}{}
+	for _, dep := range deps {
+		tMap := map[string]interface{}{}
+		// TODO: We do not save alias separate from name
```
dehume, line 135: Currently using the same subsource query used to set subsources. It seems that currently we do not store the name and alias of the tables in the system catalog. The name of the subsource will just be the `name` or the `alias` if provider. Need to think of an elegant way to handle this.

pkg/resources/resource_source_postgres.go
```diff
@@ -34,8 +35,8 @@ var sourcePostgresSchema = map[string]*schema.Schema{
 		Optional:    true,
 	},
 	"table": {
-		Description: "Creates subsources for specific tables. If neither table or schema is specified, will default to ALL TABLES",
-		Type:        schema.TypeList,
+		Description: "Creates subsources for specific tables in the Postgres connection.",
+		Type:        schema.TypeSet,
```
dehume, line 39: Has to be `TypeSet` instead of `TypeList` since we cannot control order set by the user vs order returned by the scan query.

pkg/resources/resource_source_postgres.go
```diff
@@ -85,6 +77,86 @@ func SourcePostgres() *schema.Resource {
...
+	if err := d.Set("comment", s.Comment.String); err != nil {
+		return diag.FromErr(err)
+	}
+
+	deps, err := materialize.ListDependencies(meta.(*sqlx.DB), utils.ExtractId(i), "source")
+	if err != nil {
+		return diag.FromErr(err)
+	}
+
+	// TODO: We need to use a query that retrieves name and alias
```
dehume, line 131: Currently the table information (`name` and `alias`) is not stored in the mz catalog. When this is added, will need to update and enable this section to use that.

pkg/provider/acceptance_source_postgres_test.go
```diff
@@ -36,10 +36,10 @@ func TestAccSourcePostgres_basic(t *testing.T) {
 					resource.TestCheckResourceAttr("materialize_source_postgres.test", "size", "3xsmall"),
 					resource.TestCheckResourceAttr("materialize_source_postgres.test", "text_columns.#", "1"),
 					resource.TestCheckResourceAttr("materialize_source_postgres.test", "table.#", "2"),
-					resource.TestCheckResourceAttr("materialize_source_postgres.test", "table.0.name", "table1"),
-					resource.TestCheckResourceAttr("materialize_source_postgres.test", "table.0.alias", fmt.Sprintf(`%s_table1`, connName)),
-					resource.TestCheckResourceAttr("materialize_source_postgres.test", "table.1.name", "table2"),
-					resource.TestCheckResourceAttr("materialize_source_postgres.test", "table.1.alias", fmt.Sprintf(`%s_table2`, connName)),
+					// resource.TestCheckTypeSetElemAttr("materialize_source_postgres.test", "table.0.name.*", "table1"),
```
dehume, line 39: `TestCheckTypeSetElemAttr` seems more appropriate (though this still isn't set quite right). Since switching this to `TypeSet`.

bobbyiliev:
Closing this as I've started fresh on a new branch and a new PR:

> https://github.com/MaterializeInc/terraform-provider-materialize/pull/487

Used the changes here as the base line so this has been very helpful!

PR #431, opened 2024-01-10
Update Changelog for v0.5.0

bobbyiliev:
Version `v0.5.0` introduces significant enhancements, including a unified interface for managing global and regional resources, single authentication via app password, dynamic client allocation, and advanced provider configuration options. This guide aims to assist you in a smooth transition to this new version, ensuring you can leverage these new features effectively.

## Pre-Upgrade Checklist
Before upgrading to version `v0.5.0`, please ensure that:
1. **You are currently on version `0.4.x`**: It's essential to be on the latest release of `0.4.x` series before upgrading. Version `0.4.x` includes necessary Terraform state migrations needed for a smooth transition to `v0.5.0`.
2. **Run a Terraform Plan**: After updating to the latest `0.4.x` version, execute `terraform plan` to ensure that the state migration is successful.

### New Features
- **Unified Interface**: Manage global and regional resources more efficiently.
- **Single Authentication**: Use an app password for all operations.
- **Dynamic Client Allocation**: For various resource types.
- **Enhanced Provider Configuration**: Set default settings and optional endpoint overrides.
- **New Resources**: `materialize_app_password`, `materialize_user`, and `materialize_region`.
- **SQL Connections Across Multiple Regions**: Facilitates broader reach and connectivity.
- **Region Parameter in Resources**: Specify regions for resource creation and data retrieval.

### Breaking Changes
- **Provider Configuration**:
  - Deprecated parameters: `host`, `port`, and `user`.
  - Retained parameter: `password` (for the app password).
- **New `default_region` Configuration**: Set a default region for resource creation, which can be overridden in specific resource configurations.

### Migration Steps
1. **Update Provider Configuration**: Modify your existing Terraform configurations to align with the new structure. Remove the `host`, `port`, and `user` parameters, and ensure the `password` parameter is set with your app password.

   Example:
   ```hcl
   provider "materialize" {
     password       = var.materialize_app_password
     default_region = "aws/us-east-1" # Set the default region where resources will be created
   }
   ```

2. **Manage Multi-Region Resources**: If you're working with resources across multiple regions, set the `default_region` in the provider configuration. Override this setting in specific resource blocks using the `region` parameter, if necessary.

   Example:
   ```hcl
   resource "materialize_cluster" "cluster" {
     name   = "cluster"
     region = "aws/us-west-2" # Override default region
   }
   ```

3. **Test and Validate**: After making these changes, run `terraform plan` to ensure that your configurations are correctly set up for the new version. It's recommended to perform these steps in a staging environment before applying them to production.

The new version also introduces new resources like `materialize_app_password` and `materialize_user`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+## 0.5.0 - 2024-01-10
+
 ### Features
 * Introduced a unified interface for managing both global and regional resources.
 * Implemented single authentication using an app password for all operations.


benesch (approved):

ParkMyCar:
Should we call out here that the default cluster we use for actions like creating a materialized view has changed from "default" to "quickstart"? For users whose existing default cluster is named "default", some actions may no longer work. A fix would be to run `ALTER SYSTEM SET cluster TO 'quickstart'`, not sure if that's what we always want to suggest though

benesch:
Wait, does the Terraform provider hardcode `default` as a default cluster name in places? I thought you always had to explicitly specify the cluster name, and it was just the provider's integration test suite that was hardcoding a cluster name of `default`.

bobbyiliev:
Yes, the `default` cluster was only hardcoded in the provider's acceptance and integration tests. I think that Parker is referring to the behavior on the Materialize side.

benesch:
Right, but I think it's impossible to observe that change in behavior from the Terraform provider?

bobbyiliev:
The `cluster` parameter is not required, so in the past if you were to create a materialized view it would default to the `default` cluster on the Materialize side, eg:

```
resource "materialize_materialized_view" "simple_materialized_view" {
  name          = "simple_materialized_view"

  statement = <<SQL
SELECT
    1 AS id
SQL
}
```

So on theory after the 

It is highly unlikely that people actually manage their materialized views via terraform + people actually leaving it blank, but still possible. We could make the parameter required?

benesch:
> It is highly unlikely that people actually manage their materialized views via terraform + people actually leaving it blank, but still possible. We could make the parameter required?

Shoot! That's too bad. Terraform works best when it's 100% deterministic, and doesn't rely on defaults that change. As it stands, if you fail to configure the `cluster` in a resource, things can get all out of sync if someone runs `ALTER SYSTEM SET cluster = some_other_default` in the environment.

benesch:
I filed https://github.com/MaterializeInc/terraform-provider-materialize/issues/434 to track making `cluster` required.

Issue #434, opened 2024-01-11
Make `cluster` parameter required


benesch:
The `cluster` parameter on materialized views (and possibly also indexes, sources, and sinks) is not required, e.g.:

```terraform
resource "materialize_materialized_view" "simple_materialized_view" {
  name          = "simple_materialized_view"

  statement = <<SQL
SELECT
    1 AS id
SQL
}
```

This is problematic because the cluster that will be used is the default cluster for the environment. But this can change over time if an administrator runs `ALTER SYSTEM SET cluster = ...`! That means that the same Terraform plan can create different resources in different environments or at different times, depending on the current value of the `cluster` variable, which is a bit scary and hard to reason about.

The best thing to do is probably to mandate setting `cluster` on all Terraform resources that offer it as a field. The other option is to have the Terraform provider always default the cluster to `quickstart` when unset, but that seems decidedly worse.

Discussion: https://github.com/MaterializeInc/terraform-provider-materialize/pull/431#issuecomment-1885244328

cc @bobbyiliev

PR #435, opened 2024-01-11
Make `cluster_name` parameter required

bobbyiliev:
As discussed in #434 making the `cluster_name` parameter required for materialized views and indexes.

For sinks and sources the user is already required to either specify the size of the source/sink or if they don't specify the `size`, they have to specify a `cluster_name`, eg:

https://github.com/MaterializeInc/terraform-provider-materialize/blob/a6a57bb15f132f892169a521eda672e6d3d9ead3/pkg/resources/schema.go#L387-L397


benesch (commented):
Wow, so fast. ~~Don't we need to do this for sources and sinks too?~ Didn't read your PR description. 🤦🏽‍♂️

benesch (approved):
Thank you so much!

PR #436, opened 2024-01-11
Include query with view and materialized_view

### LINKED ISSUE #122: Include query with `view` and `materialized_view`
Currently the query that reads `view` and `materialized_view` does not include the query statement. 

[mz_materialized_views](https://materialize.com/docs/sql/system-catalog/mz_catalog/#mz_materialized_views) - `definition`
[mz_views](https://materialize.com/docs/sql/system-catalog/mz_catalog/#mz_views) - `definition`

Hard part will be correctly sanitizing them.

bobbyiliev:
The query that reads view and materialized_view did not include the query statement.

Since the mz catalog has now added `create_sql`, updating the read query to include that for both view and materialized_view.

Closes #122

### DIFF
--- a/docs/resources/materialized_view.md
+++ b/docs/resources/materialized_view.md
@@ -57,6 +57,7 @@ resource "materialize_materialized_view" "simple_materialized_view" {
 
 ### Read-Only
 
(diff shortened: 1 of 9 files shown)


benesch (approved):
Nice!

PR #438, opened 2024-01-12
Drop SIZE support for sources and sinks

### LINKED ISSUE #437: Drop `SIZE` support for sources and sinks
The next version of Materialize will no longer support SIZEs for sources and sinks: https://github.com/MaterializeInc/materialize/pull/24161.

bobbyiliev:
Closes #437


ParkMyCar (approved):
I don't know the Terraform provider super well, but conceptually this makes sense to me

bobbyiliev:
> What happens to your terraform state if you have an existing linked cluster?

Yes, this is still an open question, with the current state of the PR, this will force a recreation of the source/sink in the new cluster. It would depend on the migration of converting all linked clusters into unlinked ones on the Materialize side.

benesch:
Heads up that we had to revert the database side change for v0.83, so it won't be going out to production this week.

> Yes, this is still an open question, with the current state of the PR, this will force a recreation of the source/sink in the new cluster. It would depend on the migration of converting all linked clusters into unlinked ones on the Materialize side.

I wonder if we should do a state migration that automatically fills in the `cluster` field for existing sources and sinks. The clusters would be "leaked" if you removed the source/sink from Terraform, but that seems fine. I think that would prevent the sources/sinks from being recreated when you update the resource to use `cluster` instead of `size`?

PR #440, opened 2024-01-12
comments: Mark them as public preview

ParkMyCar:
Noticed this while reviewing another PR, [COMMENT ON](https://materialize.com/docs/sql/comment-on/) is currently in public preview


bobbyiliev (changes requested):
Ah yes! Good catch @ParkMyCar 

Note that the `/docs` dir is automatically generated based on the schema of the resources/ data sources. 

So in theory this should be a single line change as the comment schema is shared between all objects:

https://github.com/MaterializeInc/terraform-provider-materialize/blob/029fbc26385fbbf126a0c558d2e4b460c00823ed/pkg/resources/schema.go#L509

bobbyiliev (approved):

PR #450, opened 2024-01-30
Allow schema rename

### LINKED ISSUE #334: Allow Schema Rename
Allow `UpdateContext` for schema resource after `ALTER SCHEMA ... RENAME ...` is implemented.

bobbyiliev:
Implementing the schema rename by itself is straight forward, the challenging part is all of the handling of all of the dependant resources, eg all resources that use the `SchemaNameSchema`.

Closes #334

### DIFF
--- a/pkg/resources/resource_schema.go
+++ b/pkg/resources/resource_schema.go
@@ -13,7 +13,7 @@ import (
 )
 
 var schemaSchema = map[string]*schema.Schema{
-	"name":               ObjectNameSchema("schema", true, true),
+	"name":               ObjectNameSchema("schema", true, false),
 	"database_name":      DatabaseNameSchema("schema", false),
 	"qualified_sql_name": QualifiedNameSchema("schema"),
 	"comment":            CommentSchema(false),
@@ -149,6 +149,15 @@ func schemaUpdate(ctx context.Context, d *schema.ResourceData, meta interface{})
 		b := materialize.NewCommentBuilder(metaDb, o)
 
 		if err := b.Object(newComment.(string)); err != nil {
+			return diag.FromErr(err)
+		}
+	}
+
+	if d.HasChange("name") {
+		oldName, newName := d.GetChange("name")
+		o := materialize.MaterializeObject{ObjectType: "SCHEMA", Name: oldName.(string), DatabaseName: databaseName}
+		b := materialize.NewSchemaBuilder(metaDb, o)
+		if err := b.Rename(newName.(string)); err != nil {
 			return diag.FromErr(err)
 		}
 	}
--- a/pkg/materialize/schema.go
+++ b/pkg/materialize/schema.go
@@ -32,6 +32,12 @@ func (b *SchemaBuilder) Create() error {
 	return b.ddl.exec(q)
 }
 
+func (b *SchemaBuilder) Rename(newName string) error {
+	old := b.QualifiedName()
+	new := QualifiedName(newName)
+	return b.ddl.rename(old, new)
+}
+
 func (b *SchemaBuilder) Drop() error {
 	qn := b.QualifiedName()
 	return b.ddl.drop(qn)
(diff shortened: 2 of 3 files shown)


pkg/resources/resource_schema.go
```diff
@@ -13,7 +13,7 @@ import (
 )
 
 var schemaSchema = map[string]*schema.Schema{
-	"name":               ObjectNameSchema("schema", true, true),
+	"name":               ObjectNameSchema("schema", true, false),
```
bobbyiliev, line 16: The problem here is that if we change the schema name to not be ForceNew, then if someone actually wants to change the name of the schema they will not be able to do that. So there needs to be a mechanism where schema renames are allowed but also they don't break all of the objects that reference that schema.

benesch: Yeah. How does this work in other database providers, like PostgreSQL and Snowflake?

bobbyiliev: Snowflake does force recreate the dependant objects, eg:

https://github.com/Snowflake-Labs/terraform-provider-snowflake/blob/d301b200b45398a1bf1e6e4127983912ba1a52a8/pkg/resources/table.go#L27-L32

They also have some other bugs that affect the way they handle schemas:

> https://github.com/Snowflake-Labs/terraform-provider-snowflake/issues/845

The Postgres providers that I've checked do not seem to support this at all, eg. they don't have resources that depend on [schema names](https://registry.terraform.io/providers/cyrilgdn/postgresql/latest/docs/resources/postgresql_schema).

bobbyiliev: One option would be to leave the ForceNew as true and then instruct users that they can add an ignore block to the child objects, eg:

```hcl
resource "materialize_database" "db" {
  name = "db"
}
resource "materialize_schema" "schema" {
  name = "test_schema3"
  database_name = materialize_database.db.name
}

resource "materialize_table" "test_table" {
  name = "test_table"
  schema_name = materialize_schema.schema.name
  # Ignore schema name rename
  lifecycle {
    ignore_changes = [schema_name]
  }
  database_name = materialize_database.db.name
  column {
    name = "id"
    type = "int"
  }
  column {
    name = "name"
    type = "text"
  }
}
```

That way the schema rename will work as expected, and the child resources will not be recreated and after the rename, the ignore block can be removed as the schema name will be updated correctly in the state file thanks to the read func.

RobinClowers: I'm way out of my depth here, but it seems like the easy answer is to not do anything special, which means users have to add the lifecycle ignore blocks like you suggested. If we do decide to set ForceNew to false, what other mechanism would we provide for schema renames?

RobinClowers: Might be question we could ask the cloud team, maybe they have suggestions since they have a lot of collective IaC experience?

bobbyiliev: After the discussion, I would say that we can take the following plan of action:

- For the short term: Merge this PR as it is as it will still allow people to rename schemas without having to drop all of their dependent objects some manual work by using the `lifecycle { ignore_changes = [schema_name] }` syntax
- For the long term: Implement `schema_id` which can be used along side `schema_name`, I've opened a tracking issue for this as it will be a pretty significant change as 80% of all resources actually depend on the `schema_name` at the moment: https://github.com/MaterializeInc/terraform-provider-materialize/issues/472

Open to suggestions! Tagging @jubrad as `schema_id` was his idea 👏

RobinClowers (approved):

pkg/resources/resource_schema.go
bobbyiliev: @benesch if you are ok with the above ^ we could go ahead and merge this, and after that plan the larger refactor in #472

benesch: Oh yes, sorry, absolutely okay with this! It's great to have an MVP!

PR #453, opened 2024-01-31
Allow imports for SSO resources

### LINKED ISSUE #451: Improve the `sso_config` data source documentation
Currently the documentation for some the `sso_config` data source documentation is a bit vague. It needs to be extended to clarify all attributes.

### LINKED ISSUE #452: Implement `import` for all SSO configuration resources
The import for the following SSO resources is currently not implemented:

- `materialize_sso_config`
- `materialize_sso_domain`
- `materialize_sso_default_roles`
- `materialize_sso_group_mapping`

bobbyiliev:
As discussed internally the SSO configuration resources did not support imports.

Closes #452 
Closes #451

### DIFF
--- a/pkg/resources/resource_sso_domain.go
+++ b/pkg/resources/resource_sso_domain.go
@@ -8,6 +8,7 @@ import (
 	"io"
 	"log"
 	"net/http"
+	"strings"
 
 	"github.com/MaterializeInc/terraform-provider-materialize/pkg/clients"
 	"github.com/MaterializeInc/terraform-provider-materialize/pkg/utils"
@@ -24,7 +25,7 @@ var SSODomainSchema = map[string]*schema.Schema{
 	"domain": {
 		Type:        schema.TypeString,
 		Required:    true,
-		Description: "The domain name for the SSO domain configuration.",
+		Description: "The domain name for the SSO domain configuration. This domain will be used to validate the SSO configuration and needs to be unique across all SSO configurations.",
 	},
 	"validated": {
 		Type:        schema.TypeBool,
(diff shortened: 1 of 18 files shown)


pkg/resources/resource_sso_domain.go
```diff
@@ -105,16 +112,36 @@ func ssoDomainRead(ctx context.Context, d *schema.ResourceData, meta interface{}
...
+			return diag.FromErr(err)
+		}
+		// Set the domain attribute
+		if err := d.Set("domain", domainName); err != nil {
+			return diag.FromErr(err)
+		}
+	} else {
+		// Regular operation, extract sso_config_id and domain from the Terraform resource data
+		ssoConfigID = d.Get("sso_config_id").(string)
+		domainName = d.Get("domain").(string)
```
sjwiesman, line 135: If we hit this else block, what is the value of Id? Curious why it isn't relevant here?

bobbyiliev: Yes, good question! The limitation here is that Frontegg does have a separate API endpoint for creating and deleting domains, but does not have a direct API endpoint for getting the domains, so to get the list of the domains, you need to call the SSO config endpoint.

By default, if this is a regular operation, eg, creating a new domain or just refreshing the state, the domain will be available from the state: `domainName = d.Get("domain").(string)`, but when we do the import, we don't have anything in the state so we need a way to pass it along, hence the `ssoConfigID:domainName` syntax where we would extract both the domain and the sso config ID from.

RobinClowers (approved):
Looks reasonable to me, with the caveat that I have very little context!

pkg/resources/resource_sso_domain.go
benesch: That makes sense. Is this a standard trick in Terraform land? I've never seen it before, but also I've imported like less than a dozen resources ever in my life.

benesch (approved):
Thanks for knocking this out so quickly!

pkg/resources/resource_sso_domain.go
bobbyiliev: Actually now thinking about this more, I've refactored this to use custom import functions, that way the read function is simplified for regular operations, and that logic is abstracted to the import function. Seems like I overengineered this yesterday 🤦

PR #456, opened 2024-02-02
Fix SSO config SP Entity ID

bobbyiliev:


pkg/provider/provider.go
```diff
@@ -49,6 +50,12 @@ func Provider(version string) *schema.Provider {
 				DefaultFunc: schema.EnvDefaultFunc("MZ_CLOUD_ENDPOINT", "https://api.cloud.materialize.com"),
 				Description: "The endpoint for the Materialize Cloud API.",
 			},
+			"base_endpoint": {
+				Type:        schema.TypeString,
+				Optional:    true,
+				DefaultFunc: schema.EnvDefaultFunc("MZ_BASE_ENDPOINT", "https://cloud.materialize.com"),
+				Description: "The base endpoint for the Materialize.",
```
benesch, line 57:
```suggestion
				Description: "The base endpoint for Materialize.",
```

benesch (approved):
Perfect! Thank you so much, Bobby!

pkg/provider/provider.go
bobbyiliev: Ah thanks! Will merge that and it will trigger new CI run, but it can be ignored as it was already successful.

PR #460, opened 2024-02-06
Mark the disk option as deprecated

bobbyiliev:


docs/resources/sso_default_roles.md
```diff
@@ -42,5 +42,5 @@ Import is supported using the following syntax:
 
 ```shell
 # Default roles can be imported using the role id, which can be found using the `materialize_sso_config` data source:
-terraform import materialize_sso_default_roles.example_role <role_id>
+terraform import materialize_sso_default_roles.example_role <configuraiton_id>
```
benesch, line 45: Typo in “configuration”.

benesch (approved):
Perfect, thank you!

PR #464, opened 2024-02-09
Add system parameter resource

### LINKED ISSUE #462: New Resource: `system_parameter`
Add a new `system_parameter` to be able to configure different system values, eg:

```sql
ALTER SYSTEM SET variable_name TO variable_value;
```

bobbyiliev:
Adding the new `system_parameter` resource.

- [x] Tests:
    - [x] Unit tests
    - [x] Integration tests
    - [x] Acceptance tests
- [x] Data source
- [x] Resource

Closes #462


pkg/datasources/datasource_system_parameter.go
```diff
@@ -0,0 +1,89 @@
...
+func systemParameterRead(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	var diags diag.Diagnostics
+
+	metaDb, region, err := utils.GetDBClientFromMeta(meta, d)
+	if err != nil {
+		return diag.FromErr(err)
+	}
+	conn := metaDb
+
+	rows, err := conn.Query("SHOW ALL;")
```
bobbyiliev, line 52: Currently the system parameter data source gets all available parameters using `show all;`. We could change this and come up with a query to only get specific parameters if we think that this might be too much.

arusahni: Would that be a breaking change, or could it be introduced in a backwards-compatible way? Are we concerned about performance w/a `SHOW ALL;`?

bobbyiliev: Yep, it should be ok as long as we keep the same schema and only change the underlying SQL statement. 

My main concern with `show all` is that it returns a lot of information and most of it might not be that helpful for the user. Though I might be looking at it wrong, and it might actually be better to have that information available in case that a user ever needs it.

arusahni: Would it complicate debugging/logging? e.g., if we need to go through verbose logs we'd need to filter out the same wall of text for each invocation?

bobbyiliev: Nope that will not be the case, the output of the query is not added to the debug output.

Issue #454, opened 2024-01-31
Add acceptance tests for the cloud resources


bobbyiliev:
Currently, we have acceptance tests only for the database resources. The could resources do have unit and integration tests, but it will be nice to also write acceptance tests for all of the new cloud resources that have been added in `v0.5.x` and `v0.6.x`.

- [ ] datasource_region
- [ ] datasource_scim_configs
- [x] datasource_sso_config
- [x] datasource_scim_groups
- [x] resource_app_password
- [x] resource_user
- [x] resource_scim_config
- [x] resource_sso_config
- [x] resource_sso_domain
- [x] resource_sso_default_roles
- [x] resource_sso_group_mapping

PR #466, opened 2024-02-13
Add acceptance tests for app password

bobbyiliev:
Planning to split this into separate PRs to avoid making one very large PR. The idea here would be to add acceptance tests for all of the new resources, and also take the chance to refactor each resource so that it utilizes the new `frontegg` package rather than having all of the logic in the `resource` package directly.

#454

### DIFF
--- a/pkg/frontegg/app_password.go
+++ b/pkg/frontegg/app_password.go
@@ -1,0 +1,71 @@
+package frontegg
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/http"
+	"time"
+
+	"github.com/MaterializeInc/terraform-provider-materialize/pkg/clients"
+)
+
+type AppPasswordResponse struct {
+	ClientID    string    `json:"clientId"`
+	Description string    `json:"description"`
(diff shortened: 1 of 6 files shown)


pkg/frontegg/app_password.go
```diff
@@ -0,0 +1,71 @@
...
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/http"
+	"time"
+
+	"github.com/MaterializeInc/terraform-provider-materialize/pkg/clients"
+)
+
+type AppPasswordResponse struct {
```
bobbyiliev, line 14: With the SQL builder, we have a separate package called `materialize` that is purely responsible for building the SQL statements. I am planning to also do the same for the Frontegg API, that way the `frontegg` package will hold all of the logic for interacting with the Frontegg API and then those funcs will be used across the different packages like data sources, resources and tests.

Initially this all lived under the `resources` package but with the additional resources that we've added recently the resource logic is getting too noisy.

RobinClowers (approved):
Splitting out the frontegg logic seems like a good idea 👍

PR #467, opened 2024-02-13
Add new cc sizes

bobbyiliev:
Adding the new cc sizes to the allowed configuration options.

The `disk` option is already marked as deprecated and should push people towards using the new sizes before we fully disable it. As of the time being this PR allows users to use both the old sizes and the new sizes.

### DIFF
--- a/pkg/resources/config.go
+++ b/pkg/resources/config.go
@@ -21,6 +21,21 @@ var replicaSizes = []string{
 	"4xlarge",
 	"5xlarge",
 	"6xlarge",
+	"25cc",
+	"50cc",
+	"100cc",
+	"200cc",
+	"300cc",
+	"400cc",
+	"600cc",
+	"800cc",
+	"1200cc",
+	"1600cc",
+	"3200cc",
+	"6400cc",
+	"128C",
+	"256C",
+	"512C",
 }
 
 var saslMechanisms = []string{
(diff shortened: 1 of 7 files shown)


pkg/resources/config.go
```diff
@@ -21,6 +21,21 @@ var replicaSizes = []string{
 	"4xlarge",
 	"5xlarge",
 	"6xlarge",
+	"25cc",
```
RobinClowers, line 24: I noticed you replaced 3xsmall with 50cc, is 25cc even smaller? Or is 50cc just a better recommendation for production use?

RobinClowers (approved):
Couple of questions and a suggestion, but looks good

pkg/resources/resource_cluster.go
```diff
@@ -118,8 +119,13 @@ func clusterCreate(ctx context.Context, d *schema.ResourceData, meta interface{}
 			b.ReplicationFactor(&r)
 		}
 
-		if v, ok := d.GetOk("disk"); ok {
-			b.Disk(v.(bool))
+		// DISK option not supported for cluster sizes ending in cc or C because disk is always enabled
+		if strings.HasSuffix(size.(string), "cc") || strings.HasSuffix(size.(string), "C") {
+			d.Set("disk", true)
```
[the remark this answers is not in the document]
bobbyiliev: Nice catch!

pkg/resources/config.go
bobbyiliev: It is mainly for the tests, the actual cluster sizes on the dev Materialize container are actually configured to be the same:

https://github.com/MaterializeInc/terraform-provider-materialize/blob/29233a0abd848e733e9dc307efd57fa63c1d1b7d/docker-compose.yml#L10

But this is a good point, I've switched the tests to 25cc for consistency.

PR #483, opened 2024-03-03
Acceptance tests for SCIM config resource

bobbyiliev:
As per #454 adding acceptance tests for `resource_scim_config`


RobinClowers (approved):
I don't have a great grasp of how these tests work, but I think it looks good.

PR #487, opened 2024-03-05
Misname subsource updates

### LINKED ISSUE #421: Failed PostgreSQL Source update causes invalid state and won't refresh

bobbyiliev:
This PR is based on #422 which was quite behind main and needed a large rebase, so starting clean here.

Fixes #421

1. Include `table` as part of the read for the resource to prevent issues when an improper table is attempted to be added.
2. Remove the options for `FOR SCHEMAS (...)` and `FOR ALL TABLES` and require the user to explicitly set every table they would like synced with the source.

This has a breaking change for users who are using the `schema` configuration property as it is getting deprecated.

The migration guide would be as follows:

1. Ensure you have a backup of your current Terraform configuration and state file. This step is crucial in case you need to revert to the previous state.

2. Make a list of all tables within the schemas you are currently using. You will need this list to explicitly define each table in your Terraform configuration.

3. In the previous versions of the Terraform provider, table configurations allowed combining the schema name with the table name, and alias could optionally include the schema name as well. The new configuration requires these elements to be explicitly separated to enhance clarity and flexibility.

The `tables` configuration property is now required and the `schema` property is removed. The `tables` property schema has been updated as follows:

```hcl
  table {
    upstream_name        = string # Required: The name of the table in the upstream database: Previously `name`
    upstream_schema_name = string # The schema of the table in the upstream database
    name                 = string # The name of the table in Materialize: Previously `alias`
    schema_name          = string # The schema of the table in Materialize
    datatabase_name      = string # The name of the database where the table will be created in Materialize
  }
```

## Migration Steps

1. First, update the Materialize Terraform provider in your configuration. You will need to modify the version in your Terraform configuration file and run:

   ```
   terraform init -upgrade
   ```

2. Update your `materialize_source_postgres` resource configuration. You need to remove the `schema` attribute and replace it with explicit `table` blocks for each table you wish to include.

   Find the `materialize_source_postgres` block in your Terraform configuration. It will look something like this:

   ```hcl
   resource "materialize_source_postgres" "example_source_postgres_schema" {
     name               = "source_postgres_schema"
     cluster_name       = "quickstart"
     publication        = "mz_source"
     schema             = ["PUBLIC"]
     postgres_connection {
       name          = materialize_connection_postgres.postgres_connection.name
       schema_name   = materialize_connection_postgres.postgres_connection.schema_name
       database_name = materialize_connection_postgres.postgres_connection.database_name
     }
   }
   ```

   Update the configuration to comment out or remove the `schema` attribute and add `table` blocks for each table:

   ```hcl
   resource "materialize_source_postgres" "example_source_postgres_schema" {
     name         = "source_postgres_schema"
     cluster_name = "quickstart"
     publication  = "mz_source"

     # schema       = ["PUBLIC"]

     table {
       upstream_name = "table1"
     }
     table {
       upstream_name = "table2"
     }
     table {
       upstream_name = "table3"
     }

     postgres_connection {
       name          = materialize_connection_postgres.postgres_connection.name
       schema_name   = materialize_connection_postgres.postgres_connection.schema_name
       database_name = materialize_connection_postgres.postgres_connection.database_name
     }
   }
   ```

On the `table` configuration side, with the latest updates, each part must be explicitly separated. This change ensures better validation at the configuration level and aligns with the SQL standards for clarity. Here’s how you should define tables now:

```
  # New way of defining tables
  table {
    upstream_name        = "table1"                  # Required: Previously `name`
    upstream_schema_name = "shop"                    # Previously part of the `name` but now has to be explicitly defined
    name                 = "table1_alias"            # Previously `alias`
    schema_name          = "shcema_in_materialize"   # Previously part of the `alias` but now has to be explicitly defined
    database_name        = "database_in_materialize" # The database where the table will be created in Materialize
  }
```

3. Apply the changes using `terraform plan` and then `terraform apply`.

   ```
   terraform apply
   ```

  > **Note:** The `terraform apply` command should not try to modify the existing resources. If it does, it means that the configuration is not correctly updated and you need to review the changes made to the `table` configuration blocks and explicitly define the tables you want to include.

4. Monitor the output and ensure that the changes are applied without any issues. Check the Materialize instance to ensure that the sources are correctly configured and operational.

## Post-migration Steps

- Validate that all intended tables are correctly sourced in Materialize and that everything is functioning as expected.

If you encounter any issues during the migration process, refer to the backup you created to revert to the previous state and seek assistance from the Materialize support channels.


pkg/resources/resource_source_postgres.go
```diff
@@ -85,6 +77,90 @@ func SourcePostgres() *schema.Resource {
...
+	for _, dep := range deps {
+		sMap := map[string]interface{}{}
+		sMap["name"] = dep.ObjectName.String
+		sMap["schema_name"] = dep.SchemaName.String
+		sMap["database_name"] = dep.DatabaseName.String
+		sMaps = append(sMaps, sMap)
+	}
+	if err := d.Set("subsource", sMaps); err != nil {
+		return diag.FromErr(err)
+	}
```
benesch, line 159: We need to resolve the TODO here to actually fix #421, right? And for that I think we need to add a new table to Materialize like `mz_postgres_sources` that exposes the table name/alias mapping.

benesch: cc @sploiselle

bobbyiliev: It will work without the TODO fix, but on the first run it will throw: `ERROR: another subsource already refers to postgres.public.table3` if the subsource already exists as Terraform will not be aware of it and will try to create it anyway. But after that first run the state will be updated and it will work as expected, so there should be no disruption on the Materialize side, besides the annoying error that the user will get. Of course it will be much cleaner to have that TODO fixed as well.

benesch: > It will work without the TODO fix, but on the first run it will throw: `ERROR: another subsource already refers to postgres.public.table3` if the subsource already exists as Terraform will not be aware of it and will try to create it anyway. But after that first run the state will be updated and it will work as expected, so there should be no disruption on the Materialize side, besides the annoying error that the user will get. Of course it will be much cleaner to have that TODO fixed as well.

Wait, that seems problematic in a different way. If adding the new table fails, why is Terraform updating the state? I would expect `terraform apply` to refrain from updating the state with the new table until it sees `ALTER SOURCE ... ADD SUBSOURCE` succeed? 

Another possible hazard: what if I manually drop a subsource out from under Terraform? I'd expect the next `terraform refresh && terraform apply` to recreate that subsource, but I think Terraform will never realize that that subsource has gone missing?

Anyway, sorry! I'm having trouble reasoning about the correctness of this change without being able to update the `tables` property! There's also something I don't understand about how the Terraform state is getting updated that seems spooky.

Adding a table to Materialize that exposes the mapping between PostgreSQL subsource table names/aliases is something that's pretty straightforward (could do in a week or two), and I'm wondering if we ought to do it before merging this PR, to make it easier to reason about the correctness here, and to not have to give users any caveats about the behavior of adding/removing subsources using Terraform after they do the migration that this PR requires.

bobbyiliev: Indeed, it will be a much cleaner solution to have the table with the mapping between subsource table names.

Speaking of this, we should do the same for the MySQL source as well as I believe that we will be in the same situation there as well, cc @rjobanp

sploiselle:
@bobbyiliev Would a system table with the schema `(subsource GlobalId, database, schema, item)` be sufficient for PG subsources? To generate the data you would need to pass to Terraform you could then do something like:

```sql
SELECT
  mz_internal.mz_global_id_to_name(id) AS subsource,
  concat_ws('.', database, schema, item) AS reference
FROM mz_pg_subsource_references
WHERE id IN ({query to get subsources of some ID})
```

This would return something like:

subsource | reference
-----------|----------
a | mydb.schema.a
another_schema.a | mydb.another_schema.a

The subsource name will be minimally qualified for the session you're connected as (i.e. for this to work naively, would need to ensure the search path is consistent for both the query and the executor of the `CREATE` statements). If that doesn't work, it's also possible to hand roll deriving the subsource name.

Another alternative would be to pre-concatenate the reference's name within the table, but denormalizing data like that is often annoying (destructuring data is hard than imposing structure on it).

And then whatever we do here, we can copy over to MySQL. The difference there would be that we'd need a different set of columns.

benesch:
@sploiselle's `concat_ws` made me realize that the design of the `table` nested resource here is wrong! It's subject to the identifier quoting issues that we've been trying so hard to stamp out elsewhere.

The way the table resource _should_ work is like this:

```hcl
table {
    schema_name = "..."
    name = " ..."
    alias = "...."
}
```

And then this code...

https://github.com/MaterializeInc/terraform-provider-materialize/blob/8e17ed241665dfaaa4314e22ad762e3011abdbf7/pkg/materialize/source_postgres.go#L95

should become:

```
			q.WriteString(fmt.Sprintf(`%s.%s AS %s`, QuoteIdentifier(t.Schema), QuoteIdentifier(t.Name), QuoteIdentifier(t.Alias)))
```

Does that make sense? This seems like the perfect time to do this since we're already forcing a migration on users. 

And then you won't need to do any `concat_ws` shenanigans when you read the tables from the system catalog. You'll just read out the schema and table names as individual values and put them in the right place. 

In general, any place where we're constructing a `db.schema.item` name as a single string is wrong, because it's basically a guarantee that we'll mishandle identifier quoting.

----

@sploiselle I think that looks about right. A few thoughts:

  * Are you sure it makes sense and/or is possible to include `database`? PostgreSQL publications are always limited to a single database, right?
  * I think we want the table to be named `mz_postgres_source_FOO` for consistency with the rest of the system catalog. Maybe we start leaning into the fact that subsources will soon be tables and call it `mz_postgres_source_tables`? 

> Another alternative would be to pre-concatenate the reference's name within the table, but denormalizing data like that is often annoying (destructuring data is hard than imposing structure on it).

Strong agree. (In fact the whole first section of my comment is an homage to this take. 😀)

> The subsource name will be minimally qualified for the session you're connected as (i.e. for this to work naively, would need to ensure the search path is consistent for both the query and the executor of the CREATE statements). If that doesn't work, it's also possible to hand roll deriving the subsource name.

Wait, can the subsource be in a schema that's _not_ the schema of the PostgreSQL source? I think we just want the Terraform provider to join against `mz_objects` and pull out the name that way.

sploiselle:
@bobbyiliev I have a local branch with something that I think covers what we need: https://github.com/sploiselle/materialize/tree/external-reference-system-table

This introduces two system tables:
- `mz_pg_source_tables`, has ID, upstream ID, schema name, table name
- `mz_mysql_source_tables`, has ID, upstream ID, table name (MySQL doesn't meaningfully support the other types of qualification)

Let me know if this is in the ballpark of the thing you're looking for. If not, glad to iterate. If so, I can tidy this up and work on merging it.

This does require the subsource dependency inversion work to merge, so you'll need to run things locally or deploy this branch to staging.

sploiselle:
@bobbyiliev Ah, there is a bug in my commit. We do need to include the schema/database name for MySQL. Will fix that later today just in case you start tinkering with this.

bobbyiliev:
> @bobbyiliev Ah, there is a bug in my commit. We do need to include the schema/database name for MySQL. Will fix that later today just in case you start tinkering with this.

Thanks @sploiselle I was just about to test this out. Poke me whenever you have the fix and I'll give it a try!

sploiselle:
@bobbyiliev Fixed this, so should be good to test/work against.
- Changed the PG table name to `mz_postgres_source_tables` for consistency's sake
- Removed the primary source ID column from both
- Added the "schema_name" column to the MySQL version

bobbyiliev:
> @bobbyiliev Fixed this, so should be good to test/work against.
> 
> * Changed the PG table name to `mz_postgres_source_tables` for consistency's sake
> * Removed the primary source ID column from both
> * Added the "schema_name" column to the MySQL version

Thanks Sean! I've just got a chance to test this branch out, but I seem to be stuck on the following panic:

```
thread 'coordinator' panicked at src/adapter/src/catalog/builtin_table_updates.rs:393:34:
primary source is a source
```

The statement that I'm trying to test this out with is:

```sql
CREATE SOURCE mz_source
  FROM POSTGRES CONNECTION postgres_connection (PUBLICATION 'mz_source')
  FOR TABLES (table1 AS s1_table1, table2 AS s2_table1);
```

sploiselle:
@bobbyiliev So annoying, sorry! Variable shadowing bit me. I also moved these tables into the `mz_internal` schema, so they're in the same place as `mz_postgres_sources`.

bobbyiliev:
Thanks @sploiselle! Just had chance to test this out and it works well!

I just have a couple of small questions:

- Is there a specific reason why the `schema_name` and the `table_name` store the values with quotes? Eg:

```sql
select * from mz_internal.mz_postgres_source_tables;
 id  | schema_name | table_name 
-----+-------------+------------
 u5  | "public"    | "table2"
 u14 | "public"    | "table2"
 u15 | "public"    | "table1"
```
 
- Do you by any chance know in which internal table do we store the relation between sources and subsource? I want to be able to tell which subsources are related with a specific source without using `SHOW SUBSOURCES ON source_name` (as it does not accept database_name as a parameter).

sploiselle:
The quotation marks provide the most stable and least sensitive option for recording identifiers, e.g.

```
postgres=# CREATE SCHEMA "1";
postgres=# CREATE TABLE "1"." a b c" ("!!" int);
SELECT "!!" FROM "1"." a b c";
```

This way you don't need to be clever or context aware to ensure that the identifiers roundtrip.

benesch:
> This way you don't need to be clever or context aware to ensure that the identifiers roundtrip.

This doesn't quite check out for me, @sploiselle! The client doesn't necessarily want to use these identifiers as identifiers in a SQL statement, nor does it necessarily want to roundtrip them. With the Terraform provider, for example, the underlying value of the identifier is what needs to get stored in the state file, so providing them as quoted identifiers would mean the Terraform provider needs to parse out the identifier's value, instead of using the column's value directly.

Another argument with a different basis: we don't quote the value of the `name` column in `mz_databases`, `mz_schemas`, `mz_objects`, etc.

Am I missing something? Is there something special about these identifiers?

sploiselle:
@benesch I'm unfamiliar with the intricacies of Terraform. Thanks for clarifying the issue. I've fixed this at https://github.com/sploiselle/materialize/tree/external-reference-system-table, @bobbyiliev

pkg/materialize/source_postgres.go
```diff
@@ -86,24 +80,17 @@ func (b *SourcePostgresBuilder) Create() error {
...
-			q.WriteString(fmt.Sprintf(`%s AS %s`, t.Name, t.Alias))
-			if i < len(b.table)-1 {
-				q.WriteString(`, `)
-			}
+	q.WriteString(` FOR TABLES (`)
+	for i, t := range b.table {
+		if t.Alias == "" {
+			t.Alias = t.Name
+		}
+		q.WriteString(fmt.Sprintf(`%s AS %s`, t.Name, t.Alias))
```
benesch, line 88: @bobbyiliev it just struck me that I think we're mishandling special characters here. We should be splitting out separate fields for `schema_name` and `alias_schema`, and then using `QuoteString` here.

bobbyiliev: This indeed seems to be the case. We are semi-handling this only in the [AddSubsource func](https://github.com/MaterializeInc/terraform-provider-materialize/pull/487/files/d24cfe8fa1a8368675c134dccd25a3f93c62efee#diff-c687ee9b62a59bc742cf585af291cacc0ebac5ac171a43a978618648c77fdd40L118-L125).

Is your idea to introduce two separate attributes as follows:

```hcl
  table {
    name  = "some_table"
    schema_name = "public"
    alias = "alias_table_name"
    alias_schema = "public"
  }
  ```

Comment (benesch) [reply pkg/materialize/source_postgres.go:88]: Yes! Although as @sploiselle mentioned on Slack, this is our chance to adjust the naming of these attributes. We could alternatively call the attributes `upstream_name`, `upstream_schema_name`, `name`, and `schema_name`.

**Comment by bobbyiliev** [comment]
Quick update here @sploiselle @benesch, I just updated the PR to introduce the changes for the subsource dependency inversion work that has already been merged and the introduction of the `mz_postgres_sources` table.

The last step here is to introduce the `table` schema as we discussed, eg:

```hcl
  table {
    name  = "some_table"
    schema_name = "public"
    upstream_name = "alias_table_name"
    upstream_schema = "public"
  }
  ```

That way we would introduce those 2 changes as a single user facing breaking change.

It will be a bit of a legwork for users as we are deprecating the `schema` attribute so this means that they will have to go and manually define all of the tables explicitly.

**Comment by benesch** [comment]
> That way we would introduce those 2 changes as a single user facing breaking change.
> 
> It will be a bit of a legwork for users as we are deprecating the `schema` attribute so this means that they will have to go and manually define all of the tables explicitly.

Sweet! I think worthwhile and definitely best to do it all at once.

@sploiselle any thoughts on the naming here? I don't have anything immediately better than `upstream_...` to suggest.

@bobbyiliev one nit: I think it should be `upstream_schema_name` rather than `upstream_schema`.

File: pkg/resources/resource_source_postgres.go [hunk 3]
```diff
@@ -95,6 +87,94 @@ func SourcePostgres() *schema.Resource {
...
+	for _, dep := range subs {
+		sMap := map[string]interface{}{}
+		sMap["name"] = dep.ObjectName.String
+		sMap["schema_name"] = dep.SchemaName.String
+		sMap["database_name"] = dep.DatabaseName.String
+		sMaps = append(sMaps, sMap)
+	}
+	if err := d.Set("subsource", sMaps); err != nil {
+		return diag.FromErr(err)
+	}
```
benesch, line 173: @bobbyiliev just making sure I'm understanding correctly — this logic is still to be updated, right?

bobbyiliev:
> > That way we would introduce those 2 changes as a single user facing breaking change.
> > It will be a bit of a legwork for users as we are deprecating the `schema` attribute so this means that they will have to go and manually define all of the tables explicitly.
> 
> Sweet! I think worthwhile and definitely best to do it all at once.
> 
> @sploiselle any thoughts on the naming here? I don't have anything immediately better than `upstream_...` to suggest.
> 
> @bobbyiliev one nit: I think it should be `upstream_schema_name` rather than `upstream_schema`.

One option here would be to go with:

```
name
schema_name
alias
alias_schema_name
```

That will allow us to keep the existing `name` and `alias` convention that the provider has had so far so it will be a less intrusive change to existing setups. But totally open to suggestions here.

pkg/resources/resource_source_postgres.go
bobbyiliev: Yep this still needs to be updated to introduce the new schema_name attr for both the upstream and local tables.

bobbyiliev:
What we've agreed on with @sploiselle just now is:

```hcl
name                -> upstream_name
schema_name         -> upstream_schema_name
alias               -> subsource_name
alias_schema_name   -> subsource_schema_name
```

Let us know if you have any concerns about this @benesch!

benesch:
Wait, won't we feel a little silly calling these "subsources" in a quarter, when we've killed the concept of a subsource?

sploiselle:
@benesch I believe we'll need to refactor these again when we remove subsources entirely. If there's a way to make these more future-proof, please do share.

benesch:
I think there might be! I sent a cal invite for 10m during lunch tomorrow
as a reminder to try to catch you guys in person—if it works out, great! If
not I'll type something up.

bobbyiliev:
> I think there might be! I sent a cal invite for 10m during lunch tomorrow as a reminder to try to catch you guys in person—if it works out, great! If not I'll type something up.

Noticed that this got canceled, totally open to leaving this as `alias_*` and then refactoring it later on when needed once the subsource concept has been deprecated on the Materialize side. So that way we don't have to now change this from `alias` to `subsources` and later on from `subsoruces` to something else.

I think that we will have to refactor this anyway depending on the exact impact on the Materialize side once the subsources concept gets deprecated.

benesch:
>  If not I'll type something up.

Finally making good on this promise.

> I believe we'll need to refactor these again when we remove subsources entirely.

If we so desired, Terraform could translate the existing `table` nested resources to `CREATE TABLE ... FROM SOURCE` commands. So it's possible that the property names we come up with now wind up living forever.

pkg/resources/resource_source_postgres.go
```diff
@@ -95,6 +99,96 @@ func SourcePostgres() *schema.Resource {
...
```
benesch, line 187: I feel like we should just delete this `subsource` property. The `tables` property has all this and more, right?

pkg/resources/resource_source_postgres.go
```diff
@@ -38,34 +39,37 @@ var sourcePostgresSchema = map[string]*schema.Schema{
...
+					Type:        schema.TypeString,
+					Optional:    true,
+					Computed:    true,
+				},
+				"schema_name": {
+					Description: "The schema of the table in Materialize.",
 					Type:        schema.TypeString,
 					Optional:    true,
+					Computed:    true,
 				},
```
benesch, line 68: Materialize supports a `database_name` as well, doesn't it?

pkg/resources/resource_source_mysql.go
```diff
@@ -38,26 +39,49 @@ var sourceMySQLSchema = map[string]*schema.Schema{
...
-				"alias": {
-					Description: "An alias for the table, used in Materialize.",
+				"schema_name": {
+					Description: "The schema of the table in Materialize.",
 					Type:        schema.TypeString,
 					Optional:    true,
+					Computed:    true,
+					ForceNew:    true,
 				},
 			},
```
benesch, line 76: Same comment as below about a database name.

pkg/resources/resource_source_mysql.go
```diff
@@ -156,7 +185,97 @@ func sourceMySQLCreate(ctx context.Context, d *schema.ResourceData, meta any) di
...
+	for _, dep := range subs {
+		sMap := map[string]interface{}{}
+		sMap["name"] = dep.ObjectName.String
+		sMap["schema_name"] = dep.SchemaName.String
+		sMap["database_name"] = dep.DatabaseName.String
+		sMaps = append(sMaps, sMap)
+	}
+	if err := d.Set("subsource", sMaps); err != nil {
+		return diag.FromErr(err)
+	}
```
benesch, line 276: Same comment as below about whether we need this now that we have the improved `table` property.

pkg/resources/resource_source_postgres.go
bobbyiliev: Ah yes! We also came to that conclusion with Sean during the on-site! Will update that accordingly

bobbyiliev:
Thanks a lot for this review and the help here @benesch and @sploiselle!

I've updated all this accordingly based on the last feedback, also I've updated the migration guide (the description of this PR) with the new changes.

 I've also tried simulating the upgrade process on my end with various different scenarios and it has been working well with the latest changes.

So far it seems to be all working well.

pkg/resources/resource_source_postgres.go
```diff
@@ -38,42 +39,50 @@ var sourcePostgresSchema = map[string]*schema.Schema{
...
+		Required: true,
+		MinItems: 1,
 	},
 	"expose_progress": IdentifierSchema(IdentifierSchemaParams{
 		Elem:        "expose_progress",
 		Description: "The name of the progress subsource for the source. If this is not specified, the subsource will be named `<src_name>_progress`.",
 		Required:    false,
 		ForceNew:    true,
 	}),
-	"subsource":      SubsourceSchema(),
```
benesch, line 76: We should remove the `func SubsourceSchema()` from schema.go too, no?

pkg/resources/resource_source_load_generator.go
```diff
@@ -192,7 +192,7 @@ func SourceLoadgen() *schema.Resource {
 		CreateContext: sourceLoadgenCreate,
```
benesch, line 192: This should lose the `subsource` property, I think!

benesch (approved):
LGTM, with one or two nits. Looks like the webhook source has a `subsource` output that should be removed too.

benesch:
Migration guide looks great, too. Once this is released we should be sure to flag to the two major customers we have using this that this will be a more involved version upgrade than usual.

pkg/resources/resource_source_postgres.go
bobbyiliev: Good point, just removed that for the rest of the sources as well.

benesch (approved):
LGTM!

PR #488, opened 2024-03-06
Add region prefix hint message

bobbyiliev:


SangJunBak (approved):
lgtm

sjwiesman (changes requested):
This message feels a little to specific to this error. Could we instead print out all the enabled regions? I assume they are the keys of the map?

bobbyiliev:
> This message feels a little to specific to this error. Could we instead print out all the enabled regions? I assume they are the keys of the map?

Great point! Just changed the message, so now users will get the following if using a wrong region:

```
│ Error: region not found: 'us-east-3'. Currently enabled regions: map[aws/eu-west-1:true aws/us-east-1:true]
```

benesch:
Nice! Can we format the map a little more cleanly than that? E.g., make a string that separates the keys in the map with a comma and a space.

bobbyiliev:
> Nice! Can we format the map a little more cleanly than that? E.g., make a string that separates the keys in the map with a comma and a space.

Just updated this so now the output will be:

```
Error: region not found: 'us-east-3'. Currently enabled regions: aws/us-east-1, aws/us-west-2
```

sjwiesman (approved):

PR #489, opened 2024-03-06
Add acceptance tests for SSO Config resource

bobbyiliev:
As part of #454 refactoring the SSO config resource to abstract the Frontegg logic to the Frontegg package and also adding acceptance tests.

### DIFF
--- a/pkg/frontegg/sso_config.go
+++ b/pkg/frontegg/sso_config.go
@@ -38,7 +38,7 @@ type Domain struct {
 	ID          string `json:"id"`
 	Domain      string `json:"domain"`
 	Validated   bool   `json:"validated"`
-	SsoConfigId string `json:"sso_config_id"`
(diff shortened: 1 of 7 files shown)


pkg/frontegg/sso_config.go
```diff
@@ -0,0 +1,205 @@
...
+	CreatedAt         time.Time `json:"createdAt"`
+	Domains           []Domain
+}
+
+// Domain represents the structure for SSO domain.
+type Domain struct {
+	ID          string `json:"id"`
+	Domain      string `json:"domain"`
+	Validated   bool   `json:"validated"`
+	SsoConfigId string `json:"sso_config_id"`
```
arusahni, line 41: To double check, this is intentionally snake-cased, right? The parent struct looks to be camelcased.

Issue #395, opened 2023-12-01
AWS Connection


dehume:
Support resource AWS connections (`materialize_connection_aws`)

| Attribute Name | Type  |
| --- | --- |
| `access_key_id` | `TypeString` | 
| `secret_access_key` | `TypeString` |
| `token` | `TypeString` |
| `assume_role_arn` | `TypeString` |
| `assume_role_session_name` | `TypeString` |

https://github.com/MaterializeInc/materialize/pull/23282

PR #417, opened 2023-12-18
[WIP] AWS Connection

dehume:
Fixes https://github.com/MaterializeInc/terraform-provider-materialize/issues/395

Add new resource `materialize_connection_aws`.

To Do:
- [X] Add builder
- [X] Add resource
- [X] Add documentation
- [X] Add integration tests
- [ ] Add acceptance tests

Issue #225, opened 2023-07-05
Investigate Terraform Plugin Framework


dehume:
The provider was built using the old Terraform SDK. We should look into what it would take to move to the [new plugin framework](https://developer.hashicorp.com/terraform/plugin/framework).

dehume:
[Benefits](https://developer.hashicorp.com/terraform/plugin/framework-benefits)
[Overview](https://developer.hashicorp.com/terraform/tutorials/providers-plugin-framework/providers-plugin-framework-provider-configure)
[Migration](https://developer.hashicorp.com/terraform/plugin/framework/migrating)

PR #492, opened 2024-03-08
AWS Connection

bobbyiliev:
Follow up PR #417

Fixes https://github.com/MaterializeInc/terraform-provider-materialize/issues/395


docs/resources/connection_aws.md
```diff
@@ -0,0 +1,122 @@
...
+### Nested Schema for `secret_access_key`
+
+Required:
+
+- `name` (String) The secret_access_key name.
+
+Optional:
+
+- `database_name` (String) The secret_access_key database name. Defaults to `MZ_DATABASE` environment variable if set or `materialize` if environment variable is not set.
+- `schema_name` (String) The secret_access_key schema name. Defaults to `public`.
```
RobinClowers, line 89: I don't quite follow, why does the `secret_access_key` have separate database and schema settings?

pkg/resources/resource_connection_aws.go
```diff
@@ -0,0 +1,275 @@
...
+		return diag.FromErr(err)
+	}
+
+	if err := d.Set("region", s.Region.String); err != nil {
+		return diag.FromErr(err)
+	}
+
+	// Note: Cannot set nested attributes with this SDK
+	// https://github.com/hashicorp/terraform-plugin-sdk/issues/459
+	// if err := d.Set("access_key_id.0.text", s.AccessKeyId.String); err != nil {
```
RobinClowers, line 109: Does this mean these attributes are not hooked up to anything? If so, we should probably comment them out so they don't show up in the docs?

docs/resources/connection_aws.md
RobinClowers: Oh I see, because the secret is a separate object. Makes sense.

RobinClowers (approved):

docs/resources/connection_aws.md
bobbyiliev: Yep exactly!

pkg/resources/resource_connection_aws.go
bobbyiliev: Nope, it's only in the read func, the create and update functions do use those attributes as normal. So this will mainly be a limitation if you try to import an existing AWS connection, the read func will not get those attributes during the import.

bobbyiliev: This will be easy to fix once we migrate to the new framework as part of #225 though this work is yet to be planned.

PR #499, opened 2024-03-13
[WIP] Terraform Plugin Framework Upgrade PoC

bobbyiliev:
A PoC to evaluate the work that would be required to shift from the Terraform SDKv2 to the new Terraform Plugin Framework as discussed in #225


bobbyiliev:
This is now at a state where it works out of the box for new resources, currently facing some issues when trying to simulate an upgrade for an existing resource created with the old SDKv2 to the new plugin framework setup. Still need to iron this out.

Issue #500, opened 2024-03-15
Load Generator Sources: auction/counter generator no longer supporting scale factor


bobbyiliev:
As per: https://github.com/MaterializeInc/materialize/pull/26037

PR #502, opened 2024-03-15
Remove scale factor for auction/counter

bobbyiliev:
Deprecating the scale factor for the load generator source except the tpch generator as per #500


pkg/materialize/source_load_generator.go
```diff
@@ -9,7 +9,6 @@ import (
 
 type CounterOptions struct {
 	TickInterval   string
-	ScaleFactor    float64
```
arusahni, line 12: We appear to be marking the field as deprecated, but are actually removing support for it?

def-: It was never supported. The value was just ignored, now the load generator is more strict and complains if you pass nonsensical arguments.

def- (approved):

pkg/materialize/source_load_generator.go
arusahni: Will the command still succeed if a scale factor is provided, or will it error out?

def-: Error.

arusahni: Got it. In that case I don't think we should be marking it as deprecated - we should remove it altogether. Deprecation generally implies that a feature still works, but will stop working soon.

bobbyiliev: Yep, the idea behind the deprecation warning at first is that I think this will officially land in v0.93.x so in a couple of weeks. So once it's officially there we'll go ahead and also remove the attribute from the schema.

arusahni: Got it! This makes sense then, especially since the field does nothing when this provider is aimed at current-stable Mz.

arusahni (approved):

PR #506, opened 2024-03-19
Allow region option for data sources

bobbyiliev:
As discussed internally, changing the `region` option for data sources to be optional so users could get override the default project region for data sources, eg:

```hcl
# data.materialize_source.eu-all
data "materialize_source" "eu-all" {
  region = "aws/eu-west-1"
}

# data.materialize_source.us-all
data "materialize_source" "us-all" {
  region = "aws/us-east-1"
}
```


benesch (approved):
Nice!

PR #509, opened 2024-03-21
Acceptance tests for SCIM Groups data source

bobbyiliev:


pkg/frontegg/scim_groups.go
```diff
@@ -0,0 +1,129 @@
...
+	var groupsResponse SCIMGroupsResponse
+	if err := json.NewDecoder(resp.Body).Decode(&groupsResponse); err != nil {
+		return nil, err
+	}
+
+	return groupsResponse.Groups, nil
+}
+
+// Helper function to flatten the roles data
+func FlattenRoles(roles []ScimRole) []interface{} {
```
SangJunBak, line 84: Why are some functions capitalized and others not?

SangJunBak (approved):
Code looks good to me! Some questions for curiosity's sake

pkg/frontegg/scim_groups.go
bobbyiliev: Thanks Jun! 

If a function name starts with an uppercase letter, it will be exported to other packages. Eg. in this case the `FlattenRoles` func will be exposed to the `resources` and `datasources` packages if it is ever needed.

PR #511, opened 2024-03-21
Allow `ALTER CONNECTION` Updates for `resource_connection_postgres`

### LINKED ISSUE #513: Allow `ALTER CONNECTION` Updates for `resource_connection_postgres`

bobbyiliev:
Implementing the `ALTER CONNECTION` Updates for `resource_connection_postgres` as per #368

This also lays down the groundwork for the other connection resources as well.

Closes #513


pkg/resources/schema.go
```diff
@@ -90,7 +90,7 @@ func ValidateConnectionSchema() *schema.Schema {
 	}
 }
 
-func IdentifierSchema(elem, description string, required bool) *schema.Schema {
+func IdentifierSchema(elem, description string, required, forceNew bool) *schema.Schema {
```
RobinClowers, line 93: Having two boolean values in a function signature is pretty confusing, it seems easy to confuse the order. Normally I'd vote for using a struct here, but I realize that there are a lot of calls to this function and it's main purpose is to build a struct in the first place, so I'm not sure what the best solution is.

pkg/provider/acceptance_connection_postgres_test.go
```diff
@@ -113,6 +113,62 @@ func TestAccConnPostgres_disappears(t *testing.T) {
 	})
 }
 
+func TestAccConnPostgres_updateConnectionAttributes(t *testing.T) {
+	initialSecretName := acctest.RandStringFromCharSet(10, acctest.CharSetAlpha)
+	updatedSecretName := acctest.RandStringFromCharSet(10, acctest.CharSetAlpha)
+	connectionName := acctest.RandStringFromCharSet(10, acctest.CharSetAlpha)
```
RobinClowers, line 119: Probably `initialConnectionName` for consistency.

RobinClowers (approved):

pkg/resources/schema.go
bobbyiliev: Good idea! I just updated this to introduce a `IdentifierSchemaParams` struct.

Let me know what you think.

RobinClowers (approved):
Nice!

PR #522, opened 2024-03-22
Add role parameter resource

### LINKED ISSUE #510: Support `alter_role_variables`
Add support for managing role specific variables, eg:

```
ALTER ROLE some_role SET transaction_isolation = 'strict serializable';
ALTER ROLE some_rple SET cluster = 'some_cluster';
```

https://materialize.com/docs/sql/alter-role/

bobbyiliev:
The only ticky part here is that there is no way to read/get the session variable defaults for a specific role.

Closes #510


pkg/materialize/role_parameter.go
```diff
@@ -0,0 +1,44 @@
...
+	return &RoleParameterBuilder{
+		ddl:           Builder{conn, System},
+		roleName:      roleName,
+		variableName:  variableName,
+		variableValue: variableValue,
+	}
+}
+
+func (b *RoleParameterBuilder) Set() error {
+	q := fmt.Sprintf(`ALTER ROLE "%s" SET "%s" TO '%s';`, b.roleName, b.variableName, b.variableValue)
```
arusahni, line 26: Should we be quoting these?

pkg/materialize/role_parameter.go
```diff
@@ -0,0 +1,44 @@
...
+	return b.ddl.exec(q)
+}
+
+func (b *RoleParameterBuilder) Reset() error {
+	q := fmt.Sprintf(`ALTER ROLE "%s" RESET "%s";`, b.roleName, b.variableName)
```
arusahni, line 31: Same question re: quoting.

pkg/materialize/role_parameter.go
```diff
@@ -0,0 +1,44 @@
...
+	return b.ddl.exec(q)
+}
+
+// TODO: Once possible, implement ShowRoleParameter
+func ShowRoleParameter(conn *sqlx.DB, roleName, variableName string) (string, error) {
+	var variableValue string
+	query := fmt.Sprintf(`SHOW "%s";`, variableName)
```
arusahni, line 38: Same question re: quoting.

arusahni (approved):
Quoting questions aside, lgtm!

pkg/materialize/role_parameter.go
bobbyiliev: Yep, I think so, do you believe that the quoting here is unnecessary? 

The idea here is that in case that a role has a special character like a space, it will not work without the quotes:

```sql
ALTER ROLE "some role" SET "cluster" = 'some cluster';
```

arusahni: Oh, I was wondering about using the `QuoteString` function instead.

bobbyiliev: Ah yes 🤦! Good catch, thanks Aru! 🙌 

Just updated this, let me know what you think.

arusahni (approved):
:sparkles:

PR #523, opened 2024-03-27
Allow alter connection params for SSH tunnels

### LINKED ISSUE #518: Allow `ALTER CONNECTION` Updates for `resource_connection_ssh_tunnel`

bobbyiliev:
As part of #368 closes #518

Also taking this as an opportunity to add validation as an arg as it seems like it has been missing for a while.

### DIFF
--- a/pkg/provider/acceptance_connection_ssh_tunnel_test.go
+++ b/pkg/provider/acceptance_connection_ssh_tunnel_test.go
@@ -53,6 +53,9 @@ func TestAccConnSshTunnel_update(t *testing.T) {
 	slug := acctest.RandStringFromCharSet(5, acctest.CharSetAlpha)
 	connectionName := fmt.Sprintf("old_%s", slug)
 	newConnectionName := fmt.Sprintf("new_%s", slug)
+	newHostName := "new_ssh_host"
+	newPort := "2222"
+	newUser := "new_ssh_user"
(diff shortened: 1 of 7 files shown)


pkg/provider/acceptance_connection_ssh_tunnel_test.go
```diff
@@ -64,10 +67,13 @@ func TestAccConnSshTunnel_update(t *testing.T) {
 				Config: testAccConnSshTunnelResource(roleName, connectionName, connection2Name, "mz_system"),
 			},
 			{
-				Config: testAccConnSshTunnelResource(roleName, newConnectionName, connection2Name, roleName),
+				Config: testAccConnSshTunnelResourceUpdated(roleName, newConnectionName, newHostName, newUser, newPort, connection2Name, roleName),
```
RobinClowers, line 70: Don't we still want to test the create as well?

RobinClowers (approved):
Looks good other than my question about the create test.

pkg/provider/acceptance_connection_ssh_tunnel_test.go
bobbyiliev: The create was still running in the previous step, but I've gone ahead and refined the test functions so that it's all in 1 function now rather than two separate ones. Thanks for that suggestion!

PR #524, opened 2024-03-27
Add public psuedo role to resource grant

bobbyiliev:
It looks like that we used to be generating the following:

```
GRANT USAGE ON CLUSTER "example_cluster" TO "PUBLIC";
ERROR:  unknown role 'PUBLIC'
```

This PR checks if the role is `PUBLIC` and if so, it does not quote it.

### DIFF
--- a/pkg/materialize/privilege.go
+++ b/pkg/materialize/privilege.go
@@ -131,6 +131,10 @@ type MaterializeRole struct {
 }
 
 func (b *MaterializeRole) QualifiedName() string {
+	// If role name is PUBLIC, it should not be quoted as it is a pseudo-role
+	if b.name == "PUBLIC" {
+		return b.name
+	}
 	return QualifiedName(b.name)
 }
 
--- a/pkg/resources/resource_grant_cluster_test.go
+++ b/pkg/resources/resource_grant_cluster_test.go
@@ -51,6 +51,42 @@ func TestResourceGrantClusterCreate(t *testing.T) {
 	})
 }
 
+func TestResourceGrantClusterCreateP(t *testing.T) {
+	utils.SetDefaultRegion("aws/us-east-1")
+	r := require.New(t)
+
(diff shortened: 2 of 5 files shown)


pkg/materialize/privilege.go
```diff
@@ -131,6 +131,10 @@ type MaterializeRole struct {
 }
 
 func (b *MaterializeRole) QualifiedName() string {
+	// If role name is PUBLIC, it should not be quoted as it is a pseudo-role
+	if b.name == "PUBLIC" {
+		return b.name
+	}
 	return QualifiedName(b.name)
 }
```
bobbyiliev, line 140: This is essentially the main change, the rest is just tests.

pkg/resources/resource_grant_cluster_test.go
```diff
@@ -51,6 +51,42 @@ func TestResourceGrantClusterCreate(t *testing.T) {
...
+
+		// Query Grant Id
+		gp := `WHERE mz_clusters.name = 'materialize'`
+		testhelpers.MockClusterScan(mock, gp)
+
+		// Query Params
+		pp := `WHERE mz_clusters.id = 'u1'`
+		testhelpers.MockClusterScan(mock, pp)
+
+		if err := grantClusterCreate(context.TODO(), d, db); err != nil {
```
SangJunBak, line 80: Out of curiosity, what does `context.TODO()` do here?

SangJunBak (approved):
LGTM

benesch (approved):
LGTM, thanks!

pkg/resources/resource_grant_cluster_test.go
bobbyiliev: Thanks Jun 🙌 

The `context.TODO()` returns a non-nil, empty Context. It is just used as a placeholder.

PR #535, opened 2024-04-03
Add cloud region resource

### LINKED ISSUE #432: Resource `materialize_region`
With `v0.5.0` now available, we can start adding additional cloud resources.

Include support for `materialize_region` as a resource.

bobbyiliev:
Closes #432


pkg/clients/cloud_client.go
```diff
@@ -115,6 +116,39 @@ func (c *CloudAPIClient) GetRegionDetails(ctx context.Context, provider CloudPro
...
+	endpoint := fmt.Sprintf("%s/api/region", provider.Url)
+	emptyJSONPayload := bytes.NewBuffer([]byte("{}"))
+	req, err := http.NewRequestWithContext(ctx, http.MethodPatch, endpoint, emptyJSONPayload)
+	if err != nil {
+		return nil, fmt.Errorf("error creating request to enable region: %v", err)
+	}
+
+	req.Header.Add("Content-Type", "application/json")
+
+	resp, err := c.FronteggClient.HTTPClient.Do(req)
```
RobinClowers, line 130: Why does this use the frontegg client? This is the cloud region api, right?

pkg/clients/cloud_client.go
```diff
@@ -115,6 +116,39 @@ func (c *CloudAPIClient) GetRegionDetails(ctx context.Context, provider CloudPro
 	return &region, nil
 }
 
+// EnableRegion sends a PATCH request to enable a region and polls until the region is enabled or a timeout is reached.
```
RobinClowers, line 119: It seems like this comment describes the region resource, this function doesn't appear to do any polling?

bobbyiliev: Ah yes, I initially had it setup that way, but a better approach is to do the polling in the resource itself rather than the client lib, so I moved the `waitForRegionToBeEnabled` to the resource itself. Will update that comment accordingly!

RobinClowers (approved):

pkg/clients/cloud_client.go
bobbyiliev: Hmm, I think that this has flowed from the initial implementation during the large refactor a few months ago.

As far as I can tell, the main idea seems to have been to reuse the FronteggClient's HTTPClient which already includes the Authorization token.

I'll have to refactor this. I've created an issue to track this and will submit a PR tomorrow! https://github.com/MaterializeInc/terraform-provider-materialize/issues/543

Thanks for pointing it out!

PR #537, opened 2024-04-05
Add key value load gen source

### LINKED ISSUE #504: New Source: add LOAD GENERATOR UPSERT
https://github.com/MaterializeInc/materialize/pull/26073

bobbyiliev:
Closes #504

### DIFF
--- a/docs/resources/source_load_generator.md
+++ b/docs/resources/source_load_generator.md
@@ -35,7 +35,7 @@ resource "materialize_source_load_generator" "example_source_load_generator" {
 
 ### Required
 
-- `load_generator_type` (String) The load generator types: [AUCTION MARKETING COUNTER TPCH].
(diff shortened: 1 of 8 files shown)


docs/resources/source_load_generator.md
```diff
@@ -91,6 +92,24 @@ Optional:
 - `schema_name` (String) The expose_progress schema name. Defaults to `public`.
 
 
+<a id="nestedblock--key_value_options"></a>
+### Nested Schema for `key_value_options`
+
+Required:
+
+- `keys` (Number) The number of keys in the source.
```
SangJunBak, line 100: I know in the docs, this number must be divisible by PARTITIONS * BATCH SIZE. Does this error propagate when someone tries running the DDL and if not, wondering if we should document it here.

Source: https://materialize.com/docs/sql/create-source/load-generator/#key-value

SangJunBak (approved):
LGTM!

PR #545, opened 2024-04-17
Add scheduling option to clusters

### LINKED ISSUE #544: Add `SCHEDULE = ON REFRESH` to clusters

bobbyiliev:
This adds the new scheduling option to clusters:

```hcl
resource "materialize_cluster" "scheduling_cluster" {
  name = "scheduling_cluster"
  size = "25cc"
  scheduling {
    on_refresh                = true
    rehydration_time_estimate = "1 hour"
  }
}
```

Closes #544

### DIFF
--- a/pkg/materialize/cluster.go
+++ b/pkg/materialize/cluster.go
@@ -3,7 +3,6 @@ package materialize
 import (
 	"database/sql"
 	"fmt"
-	"log"
 	"strings"
 
 	"github.com/jmoiron/sqlx"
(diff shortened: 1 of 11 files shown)


pkg/materialize/cluster.go
```diff
@@ -28,6 +30,27 @@ func NewClusterBuilder(conn *sqlx.DB, obj MaterializeObject) *ClusterBuilder {
...
 
+type SchedulingConfig struct {
+	onRefresh               bool
+	rehydrationTimeEstimate string
+}
+
+func GetSchedulingConfig(v interface{}) SchedulingConfig {
+	if v == nil || len(v.([]interface{})) == 0 {
+		return SchedulingConfig{}
+	}
```
SangJunBak, line 41: Why not type `v` as `[]interface{}`?  I noticed that another function calls GetSchedulingConfig with a param that's of type `v interface{}` but wouldn't that code path panic here?

pkg/provider/acceptance_cluster_test.go
```diff
@@ -395,6 +417,23 @@ func testAccClusterManagedResource(
...
+	return fmt.Sprintf(`
+resource "materialize_cluster" "test_scheduling" {
+    name                    = "%s"
+    size                    = "%s"
+    scheduling {
+        on_refresh                = %s
+        rehydration_time_estimate = "%s"
+    }
+}
+`, clusterName, size, onRefreshStr, rehydrationTimeEstimate)
```
SangJunBak, line 434: Nit: Couldn't we use `%t` instead of making `onRefreshStr`?

SangJunBak (approved):
Some questions otw lgtm!

ggevay (commented):
Sorry, I don't know Go at all, so can't comment on the details.

One thing to keep in mind is that eventually we'll have other cluster schedules, and then it might be possible to specify more than one scheduling option for a cluster, e.g., something like:
```
CREATE CLUSTER c1 (
  SIZE '1',
  SCHEDULE = ON REFRESH (REHYDRATION TIME ESTIMATE = '1 hour'),
  SCHEDULE = WAKE UP ON PEEK,
);
```
I guess the current design in this PR is future-proof for something like the above, right? E.g. one could imagine later having something like
```
resource "materialize_cluster" "scheduling_cluster" {
  name = "scheduling_cluster"
  size = "25cc"
  scheduling {
    on_refresh                = true
    rehydration_time_estimate = "1 hour"
    wake_up_on_peek           = true
  }
}
```

bobbyiliev:
> One thing to keep in mind is that eventually we'll have other cluster schedules, and then it might be possible to specify more than one scheduling option for a cluster, e.g., something like:

Ah thanks for pointing that out @ggevay! I might have to refactor this a bit so that we could support multiple schedules.

bobbyiliev:
Quick update here, as discussed in Slack, I'll refactor this so that it would look as follows:

```hcl
resource "materialize_cluster" "scheduling_cluster" {
  name = "scheduling_cluster"
  size = "25cc"
  scheduling {
    on_refresh {
       enabled = true
       rehydration_time_estimate = "1 hour"
       some_other_configs_in_the_future = true
    }
    wake_up_on_peek {
      enabled = true
      some_other_configs_in_the_future = true
    }
  }
}
```

benesch:
Really like the look of that, @bobbyiliev! Nice!

ggevay:
To me, this is looking good, but I'm really not an authority on Terraform or Go. @SangJunBak, could you please take another look?

PR #546, opened 2024-04-17
Drop public schema by default for database resource

### LINKED ISSUE #284: Better handling of the `public` schema
When creating a database with Terraform, Materialize automatically creates a `public` schema in that database behind Terraform's back. It's not easily possible to manage this `public` schema with Terraform. You can't declare a resource for it, because that will fail with "resource already exists." So you have to manually import into state, which cannot (AFAIUI) be automated.

This is particularly problematic when changing the owner of the database. Say you create a database `foo` owned by `some_role`:

```tf
resource "materialize_database" "foo" {
  name           = "foo"
  ownership_role = "some_role"
}
```

The `foo.public` schema will awkwardly be owned by whatever user Terraform is running as, and `some_role` will not have any privileges on this schema.

Assigning default privileges to `some_role` also doesn't help, because those default privileges will be installed _after_ the `public` schema is created, and default privileges only apply to future objects, not existing objects.

Similar issues in other Terraform providers:

  * https://github.com/Snowflake-Labs/terraform-provider-snowflake/issues/306
  * https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/security_group#argument-reference, which automatically deletes the default `ALLOW ALL` egress rule

bobbyiliev:
As per the discussion here https://github.com/MaterializeInc/terraform-provider-materialize/issues/284 

Closes #284

---

## Migration Guide: Handling Public Schemas with Terraform in Materialize

With the latest update to our Terraform provider for Materialize, the automatic creation of a public schema within new databases has been removed. This change aligns more closely with common infrastructure-as-code practices, providing users with more control over schema management and security settings. 

Here is a quick guide that will walk you through the necessary steps to manually create and configure a public schema if you need its presence.

### Changes in Behavior
Prior to this update, Materialize would automatically create a public schema in each new database, matching traditional SQL database behaviors. However, with the new version of the Terraform provider, this is no longer the case. You now need to explicitly define and manage the public schemas within your Terraform configuration.

#### 1. Update Your Terraform Configuration

Existing databases will not be affected, the change only affects newly created databases using the `materialize_database` resource.

If your databases rely on the public schema, you will need to add a few lines to your Terraform configuration to manually create this schema and set the necessary permissions. Below is an example configuration that illustrates these changes:

```hcl
# Create a Materialize database without a public schema
resource "materialize_database" "example" {
  name = "example"
}

# Optionally create a public schema in the database
resource "materialize_schema" "public" {
  name          = "public"
  database_name = materialize_database.example.name
}

# Grant USAGE to the PUBLIC pseudo-role for the public schema
# This matches the default behavior of Materialize
resource "materialize_schema_grant" "schema_grant_usage" {
  role_name     = "PUBLIC"
  privilege     = "USAGE"
  database_name = materialize_database.example.name
  schema_name   = materialize_schema.public.name
}
```

#### 2. Apply Configuration Changes
After updating your configuration files, run `terraform apply` to apply the changes. Ensure that you review the plan before applying to avoid unintended modifications to your infrastructure.

#### 3. Integrating the Public Schema with Other Resources

Once the public schema is successfully created and configured, you can integrate it with other resources within your Materialize environment by referencing the schema name and database name:

```hcl
resource "materialize_connection_ssh_tunnel" "ssh_connection" {
  name          = "ssh_connection"
  database_name = materialize_database.example.name
  schema_name   = materialize_schema.public.name
  comment       = "Connection through SSH tunnel for secure data transfer"

  host = "ssh_host"
  user = "ssh_user"
  port = 22

  validate = false
}
```

#### 4.  Importing Existing Schemas and Grants

If you have existing public schemas that were automatically created in previous versions, and if you need to manage them in your Terraform project, you'll need to first import these resources into your new Terraform configuration to manage them effectively.

Start by first defining the resources in your project, and then:

- For schemas, you can import them into your Terraform state using the following command:

    ```shell
    terraform import materialize_schema.example_schema <region>:<schema_id>
    ```

    Here, `<region>` is the region where your database is located (e.g., `aws/us-east-1`), and `<schema_id>` is the unique identifier for the schema, which you can find in the `mz_catalog.mz_schemas` table.

- For schema grants, use the following format to ensure that existing permissions are accurately represented in your Terraform state:

    ```shell
    terraform import materialize_schema_grant.example_grant <region>:GRANT|SCHEMA|<schema_id>|<role_id>|<privilege>
    ```

This update to the Materialize Terraform provider aims for a more explicit and controlled configuration approach in database schema management.

### DIFF
--- a/docs/resources/database.md
+++ b/docs/resources/database.md
@@ -25,7 +25,7 @@ resource "materialize_database" "example" {
 # The Terraform provider on the other hand does not create a public schema by default
 # Optionally you can create a public schema in the database using the materialize_schema resource
 resource "materialize_schema" "public" {
-  name     = "public"
+  name          = "public"
   database_name = materialize_database.example.name
 }
 
--- a/docs/resources/database.md
+++ b/docs/resources/database.md
@@ -28,7 +28,6 @@ resource "materialize_database" "example_database" {
 ### Optional
 
 - `comment` (String) **Public Preview** Comment on an object in the database.
-- `no_public_schema` (Boolean) When set to true, the default 'public' schema will be dropped from the database.
 - `ownership_role` (String) The owernship role of the object.
 - `region` (String) The region to use for the resource connection. If not set, the default region is used.
 
--- a/docs/resources/database.md
+++ b/docs/resources/database.md
@@ -4,17 +4,38 @@ page_title: "materialize_database Resource - terraform-provider-materialize"
 subcategory: ""
 description: |-
   The highest level namespace hierarchy in Materialize.
+  Note: This resource will not automatically create a public schema.If needed, the public schema must be explicitly defined in your configuration using the materialize_schema resource.
 ---
 
 # materialize_database (Resource)
 
 The highest level namespace hierarchy in Materialize.
 
+**Note**: This resource will not automatically create a public schema.If needed, the public schema must be explicitly defined in your configuration using the `materialize_schema` resource.
+
 ## Example Usage
 
 ```terraform
-resource "materialize_database" "example_database" {
-  name = "database"
+# Create a Materialize database without a public schema
+resource "materialize_database" "example" {
+  name = "example"
+}
+
+# By default, Materialize creates a public schema in each database
+# The Terraform provider on the other hand does not create a public schema by default
+# Optionally you can create a public schema in the database using the materialize_schema resource
+resource "materialize_schema" "public" {
+  name     = "public"
+  database_name = materialize_database.example.name
+}
+
+# Grant USAGE to the PUBLIC pseudo-role for the public schema
+# This matches the default behavior of Materialize
+resource "materialize_schema_grant" "schema_grant_usage" {
+  role_name     = "PUBLIC"
+  privilege     = "USAGE"
+  database_name = materialize_database.example.name
+  schema_name   = materialize_schema.public.name
 }
 ```
 
--- a/docs/resources/database.md
+++ b/docs/resources/database.md
@@ -28,6 +28,7 @@ resource "materialize_database" "example_database" {
 ### Optional
 
 - `comment` (String) **Public Preview** Comment on an object in the database.
+- `no_public_schema` (Boolean) When set to true, the default 'public' schema will be dropped from the database.
 - `ownership_role` (String) The owernship role of the object.
 - `region` (String) The region to use for the resource connection. If not set, the default region is used.
 
--- a/docs/resources/scim_group_roles.md
+++ b/docs/resources/scim_group_roles.md
@@ -3,7 +3,7 @@
 page_title: "materialize_scim_group_roles Resource - terraform-provider-materialize"
 subcategory: ""
 description: |-
-  The materializescimgroup_role resource allows managing roles within a SCIM group in Frontegg.
+  The materialize_scim_group_role resource allows managing roles within a SCIM group in Frontegg.
 ---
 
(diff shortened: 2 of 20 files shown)


docs/resources/database.md
```diff
@@ -28,6 +28,7 @@ resource "materialize_database" "example_database" {
 ### Optional
 
 - `comment` (String) **Public Preview** Comment on an object in the database.
+- `no_public_schema` (Boolean) When set to true, the default 'public' schema will be dropped from the database.
```
benesch, line 31: I'd flip this around and make it `public_schema` or `create_public_schema` — in general I think it's much easier to reason about things when all boolean parameters are positive rather than negative.

benesch (commented):
Spicy thought to suggest: what if we didn't add the option at all? What if Terraform _always_ dropped the public schema when creating a fresh database? That fits a lot better with Terraforms model. If you want the public schema, you need to create it with a `materialize_schema` resource.

benesch:
> Spicy thought to suggest: what if we didn't add the option at all? What if Terraform _always_ dropped the public schema when creating a fresh database? That fits a lot better with Terraforms model. If you want the public schema, you need to create it with a `materialize_schema` resource.

Another benefit of doing it this way: you don't have to handle the case where `include_public_schema = false` is changed to `include_public_schema = true`. Like, setting that parameter to true after it's previously been false _should_ entail recreating the public schema ... but that's pretty awkward to code up in Terraform.

bobbyiliev:
> Spicy thought to suggest: what if we didn't add the option at all? What if Terraform _always_ dropped the public schema when creating a fresh database? That fits a lot better with Terraforms model. If you want the public schema, you need to create it with a `materialize_schema` resource.

I quite like this actually! Will adjust the PR today.

benesch (approved):
LGTM! Just need to write up a big migration guide that shows users how to create the `public` schema if they want it. Should also probably add a big note to the `materialize_database` resource indicating that this is a deviation from how databases work in the SQL interface. We should show users in the docs the exact resources that they can copy and paste to create the `public` schema. Note that it's not just the `materialize_schema` resource—IIUC, the `public` schema also has privileges attached that allow the `PUBLIC` pseudo-role `USAGE` access.

docs/resources/scim_group_roles.md
```diff
@@ -3,7 +3,7 @@
 page_title: "materialize_scim_group_roles Resource - terraform-provider-materialize"
 subcategory: ""
 description: |-
-  The materializescimgroup_role resource allows managing roles within a SCIM group in Frontegg.
+  The materialize_scim_group_role resource allows managing roles within a SCIM group in Frontegg.
```
bobbyiliev, line 6: Some of those docs changes were generated after this Dependabot PR #549

pkg/resources/resource_database.go
```diff
@@ -85,6 +85,11 @@ func databaseCreate(ctx context.Context, d *schema.ResourceData, meta interface{
 		return diag.FromErr(err)
 	}
 
+	// drop public schema by default
+	if err := b.DropPublicSchema(); err != nil {
+		return diag.FromErr(err)
+	}
+
```
bobbyiliev, line 92: This is the core change. The rest is mainly documentation and test updates.

benesch (approved):
This looks fantastic, Bobby, thank you! I think we can say that this "Closes #284."

I have one request re: notifying our existing customers about this change. I'll sync with you on that in Slack.

benesch (approved):

PR #550, opened 2024-04-23
Fix resource imports for non-default regions

bobbyiliev:


pkg/resources/resource_grant_role.go
```diff
@@ -85,9 +85,10 @@ func grantRoleRead(ctx context.Context, d *schema.ResourceData, meta interface{}
 	// Check if role contains member
 	mapping, _ := materialize.ParseRolePrivileges(roles)
 
+	// Check if role contains member and if not, set id to empty string so that it can be recreated
 	if !slices.Contains(mapping[key.roleId], key.memberId) {
 		d.SetId("")
-		return diag.Errorf("role does contain member %s", key.memberId)
+		return nil
 	}
```
bobbyiliev, line 92: There was an issue with the logical here when handling the situation where a role does not contain a member. The code checks whether the specified member is part of the role, and if not, it sets the resource ID to an empty string `(d.SetId(""))` which effectively tells Terraform that the resource does not exist. 

However, instead of just setting the ID to an empty string and returning `nil` which would indicate a successful "non-existence" handling to Terraform, the old code returned an error after setting the ID to an empty string which essentially never removed the non-existing resource from the state.

pkg/utils/provider_meta.go
```diff
@@ -59,6 +59,8 @@ func GetDBClientFromMeta(meta interface{}, d *schema.ResourceData) (*sqlx.DB, cl
 	var region clients.Region
 	if d != nil && d.Get("region") != "" {
 		region = clients.Region(d.Get("region").(string))
+	} else if d != nil && ExtractRegion(d.Id()) != "" {
+		region = clients.Region(ExtractRegion(d.Id()))
 	} else {
 		region = providerMeta.DefaultRegion
 	}
```
bobbyiliev, line 66: There was a bug here which only manifested when importing resources with a non-default region.

During the import stage, we essentially only have access to the ID of the specific resource, eg:

```
terraform import materialize_role.example_role <region>:<role_id>
```

However we never checked the old logic was only checking for the `region` attribute which is not available during the import stage, and else it was using the default region, which might have been different from the region of the imported resource.

RobinClowers (approved):
Makes sense, just one suggestion!

pkg/utils/provider_meta.go
```diff
@@ -110,3 +112,13 @@ func ExtractId(oldID string) string {
...
 	return parts[1]
 }
+
+// Function to get the region from the region + ID string
+func ExtractRegion(oldID string) string {
+	parts := strings.Split(oldID, ":")
+	if len(parts) < 2 {
+		// Return an empty string if the ID doesn't have a region
+		return ""
```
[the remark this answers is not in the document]
bobbyiliev: Unfortunately, in Go, you cannot return `nil` for a value that is expected to be a `string`. It would error out with:

```
cannot use nil as string value in return statement compiler
```

PR #552, opened 2024-04-26
Check for nil values in GetSliceValueString

### LINKED ISSUE #555: Fix: panic: interface conversion: interface {} is nil, not string

bobbyiliev:
As reported internally, this fixes a panic when a nil value is passed to `GetSliceValueString`.

Closes #555

```
panic: interface conversion: interface {} is nil, not string
```

### DIFF
--- a/pkg/materialize/utils.go
+++ b/pkg/materialize/utils.go
@@ -26,7 +26,12 @@ func QualifiedName(fields ...string) string {
 func GetSliceValueString(v []interface{}) []string {
 	var o []string
 	for _, i := range v {
-		o = append(o, i.(string))
+		if i != nil {
+			str, ok := i.(string)
+			if ok {
+				o = append(o, str)
+			}
+		}
 	}
 	return o
 }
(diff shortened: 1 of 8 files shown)


pkg/materialize/utils.go
```diff
@@ -26,7 +26,12 @@ func QualifiedName(fields ...string) string {
 func GetSliceValueString(v []interface{}) []string {
 	var o []string
 	for _, i := range v {
-		o = append(o, i.(string))
+		if i != nil {
+			str, ok := i.(string)
+			if ok {
+				o = append(o, str)
+			}
+		}
```
benesch, line 34: I think we should return an error here rather than silently excluding nil! E.g., IMO it is just a straight error to specify `text_columns = [nil]`, and all we need to do is return a helpful error message rather than panicking. Does that check out for you?

bobbyiliev: Good idea! Just refactored the PR to return an error instead.

benesch (approved):
Nice, thanks.

RobinClowers (approved):

PR #561, opened 2024-04-30
Fix Kafka connection rename race condition

bobbyiliev:
Moving the Kafka connection rename statement to be executed first in order to prevent a race condition when updating multiple parameters at once.


docs/data-sources/scim_groups.md
```diff
@@ -38,7 +38,7 @@ Read-Only:
 - `users` (List of Object) (see [below for nested schema](#nestedobjatt--groups--users))
 
 <a id="nestedobjatt--groups--roles"></a>
-### Nested Schema for ``
+### Nested Schema for `groups.roles`
```
bobbyiliev, line 41: Those docs changes are triggered due to the latest Dependabot PR #558

PR #573, opened 2024-06-03
Allow skipping activation emails when creating users

benesch:
This is useful for customers who need to create service accounts and don't have a dedicated email address for such accounts.

This is a partial solution to #20894.


ParkMyCar (approved):

bobbyiliev (approved):
Looks good! 👏 

I will test this shortly and merge it later on today

PR #578, opened 2024-06-11
Allow Kafka Sink From attribute updates

### LINKED ISSUE #571: Kafka Sink Resource: implement alter sink
https://github.com/MaterializeInc/materialize/pull/27379

bobbyiliev:
Fixes #571

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -39,7 +39,7 @@ services:
 
   redpanda:
     container_name: redpanda
-    image: docker.vectorized.io/vectorized/redpanda:v21.11.2
+    image: docker.vectorized.io/vectorized/redpanda:v23.3.5
     depends_on:
       - postgres
     command:
--- a/pkg/materialize/generic.go
+++ b/pkg/materialize/generic.go
@@ -161,3 +161,8 @@ func (b *Builder) alterDrop(name string, options []string, validate bool) error 
 
 	return b.exec(query)
 }
+
+func (b *Builder) alterFrom(name string, from IdentifierSchemaStruct) error {
+	query := fmt.Sprintf(`ALTER %s %s SET FROM %s;`, b.entity, name, from.QualifiedName())
+	return b.exec(query)
+}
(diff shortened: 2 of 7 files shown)


docker-compose.yml
```diff
@@ -39,7 +39,7 @@ services:
 
   redpanda:
     container_name: redpanda
-    image: docker.vectorized.io/vectorized/redpanda:v21.11.2
+    image: docker.vectorized.io/vectorized/redpanda:v23.3.5
```
bobbyiliev, line 42: Taking the chance to upgrade the Redpanda instance.

pkg/resources/resource_sink_kafka.go
```diff
@@ -24,7 +24,7 @@ var sinkKafkaSchema = map[string]*schema.Schema{
 		Elem:        "from",
 		Description: "The name of the source, table or materialized view you want to send to the sink.",
 		Required:    true,
-		ForceNew:    true,
+		ForceNew:    false,
```
bobbyiliev, line 27: By setting the `ForceNew` flag to false, we will no longer have to drop and re-create the sink if the `from` attribute changes.

pkg/materialize/generic.go
```diff
@@ -161,3 +161,8 @@ func (b *Builder) alterDrop(name string, options []string, validate bool) error
 
 	return b.exec(query)
 }
+
+func (b *Builder) alterFrom(name string, from IdentifierSchemaStruct) error {
+	query := fmt.Sprintf(`ALTER %s %s SET FROM %s;`, b.entity, name, from.QualifiedName())
```
RobinClowers, line 166: Does `QualifiedName` handle escaping the input?

RobinClowers (approved):

pkg/materialize/generic.go
bobbyiliev: Yep!

```go
func QuoteIdentifier(input string) string {
	return `"` + strings.Replace(input, `"`, `""`, -1) + `"`
}

func QualifiedName(fields ...string) string {
	var o []string
	for _, f := range fields {
		c := QuoteIdentifier(f)
		o = append(o, c)
	}

	q := strings.Join(o[:], ".")
	return q
}
```

PR #595, opened 2024-07-08
Only fetch the list of Frontegg roles once per TF provider invocation

### LINKED ISSUE #579: Only fetch the list of SSO roles once per TF provider invocation
At the moment we fetch the list of SSO roles ad-hoc for every resource. It will be beneficial to refactor this so that we only do this once during the provider invocation and then use that throughout the different resources/data sources.

bobbyiliev:
Fixes #579

In the past we were making requests to fetch the Frontegg roles for each individual resource, with this refactor we would only make that request and store the roles data in the newly introduced `FronteggRoles` provider metadata attribute.


RobinClowers (approved):
Nice!

PR #603, opened 2024-07-12
Change REHYDRATION TIME ESTIMATE to HYDRATION TIME ESTIMATE

### LINKED ISSUE #601: Change `REHYDRATION TIME ESTIMATE` to `HYDRATION TIME ESTIMATE`
As per this change here: https://github.com/MaterializeInc/materialize/pull/28138

bobbyiliev:
Fixes #601 

As per @ggevay'scomment `REHYDRATION TIME ESTIMATE` will still be accepted by the parser so I am marking the `rehydration_time_estimate` as deprecated and will remove it once the parser no longer accepts it:
> This changes the REHYDRATION TIME ESTIMATE cluster option to HYDRATION TIME ESTIMATE. The parser still accepts the old version. Eventually, the old version will be completely removed, once users move to the new version.

### DIFF
--- a/docs/resources/cluster.md
+++ b/docs/resources/cluster.md
@@ -55,7 +55,8 @@ Optional:
 Optional:
 
 - `enabled` (Boolean) Enable scheduling to refresh the cluster.
-- `rehydration_time_estimate` (String) Estimated time to rehydrate the cluster during refresh.
+- `hydration_time_estimate` (String) Estimated time to hydrate the cluster during refresh.
+- `rehydration_time_estimate` (String, Deprecated) Estimated time to rehydrate the cluster during refresh.
 
 ## Import
 
--- a/pkg/materialize/cluster.go
+++ b/pkg/materialize/cluster.go
@@ -35,6 +35,7 @@ type SchedulingConfig struct {
 
 type OnRefreshConfig struct {
 	Enabled                 bool
+	HydrationTimeEstimate   string
 	RehydrationTimeEstimate string
 }
 
@@ -66,6 +67,7 @@ func GetSchedulingConfig(v interface{}) SchedulingConfig {
 	return SchedulingConfig{
 		OnRefresh: OnRefreshConfig{
 			Enabled:                 onRefreshMap["enabled"].(bool),
+			HydrationTimeEstimate:   onRefreshMap["hydration_time_estimate"].(string),
 			RehydrationTimeEstimate: onRefreshMap["rehydration_time_estimate"].(string),
 		},
 	}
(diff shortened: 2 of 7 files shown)


RobinClowers (approved):

docs/resources/cluster.md
```diff
@@ -55,7 +55,8 @@ Optional:
 Optional:
 
 - `enabled` (Boolean) Enable scheduling to refresh the cluster.
-- `rehydration_time_estimate` (String) Estimated time to rehydrate the cluster during refresh.
+- `hydration_time_estimate` (String) Estimated time to hydrate the cluster during refresh.
+- `rehydration_time_estimate` (String, Deprecated) Estimated time to rehydrate the cluster during refresh.
```
ggevay, line 59: Maybe mention that both options do exactly the same thing.

pkg/materialize/cluster.go
```diff
@@ -151,7 +165,9 @@ func (b *ClusterBuilder) Create() error {
 
 		if b.schedulingConfig.OnRefresh.Enabled {
 			scheduleStatement := " SCHEDULE = ON REFRESH"
-			if b.schedulingConfig.OnRefresh.RehydrationTimeEstimate != "" {
+			if b.schedulingConfig.OnRefresh.HydrationTimeEstimate != "" {
+				scheduleStatement += fmt.Sprintf(" (HYDRATION TIME ESTIMATE = %s)", QuoteString(b.schedulingConfig.OnRefresh.HydrationTimeEstimate))
+			} else if b.schedulingConfig.OnRefresh.RehydrationTimeEstimate != "" {
 				scheduleStatement += fmt.Sprintf(" (REHYDRATION TIME ESTIMATE = %s)", QuoteString(b.schedulingConfig.OnRefresh.RehydrationTimeEstimate))
```
ggevay, line 171: Would it be possible to make Terraform write "HYDRATION TIME ESTIMATE" in the SQL, even when the user used the `RehydrationTimeEstimate` option at the Terraform level? This way, even if some Terraform users don't move to the new option name at the Terraform level, we can still remove the old syntax from SQL, as soon as they all simply move to the new version of Terraform.

pkg/materialize/cluster.go
```diff
@@ -214,7 +230,9 @@ func (b *ClusterBuilder) SetSchedulingConfig(s interface{}) error {
 	// Check if the scheduling is enabled and set the appropriate SQL command.
 	if schedulingConfig.OnRefresh.Enabled {
 		scheduleStatement = "SCHEDULE = ON REFRESH"
-		if schedulingConfig.OnRefresh.RehydrationTimeEstimate != "" {
+		if schedulingConfig.OnRefresh.HydrationTimeEstimate != "" {
+			scheduleStatement += fmt.Sprintf(" (HYDRATION TIME ESTIMATE = %s)", QuoteString(schedulingConfig.OnRefresh.HydrationTimeEstimate))
+		} else if schedulingConfig.OnRefresh.RehydrationTimeEstimate != "" {
 			scheduleStatement += fmt.Sprintf(" (REHYDRATION TIME ESTIMATE = %s)", QuoteString(schedulingConfig.OnRefresh.RehydrationTimeEstimate))
```
ggevay, line 236: (And same here I guess.)

pkg/materialize/cluster.go
bobbyiliev: Good idea! Will change that now!

ggevay (approved):
Thank you!

PR #604, opened 2024-07-12
Add support for adding and removing MySQL subsources

### LINKED ISSUE #490: MySQL Source: Add `ALTER SOURCE...ADD SUBSOURCES` once supported by Materialize

bobbyiliev:
Fixes #490


pkg/resources/resource_source_mysql.go
```diff
@@ -327,3 +341,20 @@ func sourceMySQLDelete(ctx context.Context, d *schema.ResourceData, meta any) di
...
+
+	var difference []string
+	for _, item := range arr1 {
+		i := item.(string)
+		if !arr2Map[i] {
+			difference = append(difference, i)
+		}
+	}
+	return difference
+}
```
SangJunBak, line 360: It doesn't seem like we use this anywhere!

SangJunBak (approved):
LGTM! Small nit tho

pkg/resources/resource_source_mysql.go
bobbyiliev: Ah yes! Good catch! 🙏

PR #612, opened 2024-07-17
Add support for upsert error decoding alias

### LINKED ISSUE #605: Add new upsert sources option: Allow choosing the name of the errors column
https://github.com/MaterializeInc/materialize/issues/28258

bobbyiliev:
Fixes #605

CI tests will be failing until the new image has been released.

The initial implementation of the value decoding errors attribute in the Terraform schema was just a string and was defined as follows:

```hcl
  envelope {
    upsert = true
    upsert_options {
      value_decoding_errors = "INLINE"
    }
  }
```

However as the `inline` attribute will now have some extra configuration options itself it would be cleaner to have the following instead:

```hcl
  envelope {
    upsert = true
    upsert_options {
      value_decoding_errors {
        inline {
          enabled = true
          alias   = "my_error_col"
        }
      }
    }
  }
```

This will give us the ability to add more configuration options to `INLINE` and also add additional value decoding error methods.

UPDATE: We can discard the bellow as we've agreed to remove the schema migration:

The main thing with this change is that if a user created an upsort with `value_decoding_errors` with a previous version of the provider, this would have stored the value as string in their terraform state file, eg:

```json
"envelope": [
  {
    "debezium": false,
    "none": false,
    "upsert": true,
    "upsert_options": [
      {
        "value_decoding_errors": "INLINE"
      }
    ]
  }
],
```

This would break any terraform run with `Planning failed. Terraform encountered an error while generating this plan. Error: missing expected [`. 

To handle this I've added a state migration to transform the state file to the new format, eg:

```json
"envelope": [
  {
    "debezium": false,
    "none": false,
    "upsert": true,
    "upsert_options": [
      {
        "value_decoding_errors": [
          {
            "inline": [
              {
                "alias": "",
                "enabled": true
              }
            ]
          }
        ]
      }
    ]
  }
],
```

With that state migration in place, users should be able to run `terraform init -upgrade` to get the new provider version, and then `terraform plan` and `terraform apply` should work without errors, automatically converting the old state format to the new one.

As a side note, we should be able to remove the state migration if we knew that no one is using `value_decoding_errors` just yet.


docs/resources/source_kafka.md
```diff
@@ -111,7 +111,24 @@ Optional:
...
+Optional:
+
+- `inline` (Block List, Max: 1) Configuration for inline value decoding errors. (see [below for nested schema](#nestedblock--envelope--upsert_options--value_decoding_errors--inline))
+
+<a id="nestedblock--envelope--upsert_options--value_decoding_errors--inline"></a>
+### Nested Schema for `envelope.upsert_options.value_decoding_errors.inline`
+
+Optional:
+
+- `alias` (String) Specify an alias for the value decoding errors column, to use an alternative name for the error column.
```
rjobanp, line 128:
```suggestion
- `alias` (String) Specify an alias for the value decoding errors column, to use an alternative name for the error column. If not specified the column name will be `error`
```

rjobanp (approved):

PR #614, opened 2024-07-19
Extend integration tests to run in two regions

### LINKED ISSUE #560: Add tests for multiple regions
At the moment the way the tests are setup, they only default to a single region. With the recent few bugs reported when using multiple regions, it will be nice to extend the tests so that we simulate multiple regions.

bobbyiliev:
Fixes #560

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -42,10 +42,10 @@ services:
       - --system-parameter-default=max_aws_privatelink_connections=10
       - --all-features
     environment:
-      MZ_SQL_LISTEN_ADDR: 7875
-      MZ_HTTP_LISTEN_ADDR: 7876
-      MZ_INTERNAL_SQL_LISTEN_ADDR: 7877
-      MZ_INTERNAL_HTTP_LISTEN_ADDR: 7878
+      MZ_SQL_LISTEN_ADDR: 0.0.0.0:7875
+      MZ_HTTP_LISTEN_ADDR: 0.0.0.0:7876
+      MZ_INTERNAL_SQL_LISTEN_ADDR: 0.0.0.0:7877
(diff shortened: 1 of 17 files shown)


SangJunBak (approved):
Code looks good to me! Unfortunate that users with multiple regions will have to duplicate a lot of their objects, but doesn't seem like there's a terraform-only solution for this.

benesch:
> Unfortunate that users with multiple regions will have to duplicate a lot of their objects, but doesn't seem like there's a terraform-only solution for this.

If they're looking to create the same objects in multiple regions, they can use Terraform's take on loops (`count`) to automatically create the same objects in multiple regions—is that the kind of thing you meant?

bobbyiliev:
Indeed, this can be done with a loop:

```hcl
variable "regions" {
  type    = list(string)
  default = ["aws/us-west-2", "aws/us-east-1"]
}

resource "materialize_cluster" "multi_region_clusters" {
  count   = length(var.regions)
  name    = "cluster"
  comment = "Cluster in ${var.regions[count.index]}"
  region  = var.regions[count.index]
}
```

My idea behind the test here is to make sure that users can rely on the default region in their provider configuration, but also be able to spin up ad-hoc resources in different regions by overriding the default one in the resources themselves. 

I think that in most cases, people only manage resources in 1 region, and if they wanted to replicate their whole setup in another region they would just change the default region provider configuration value.

SangJunBak:
>If they're looking to create the same objects in multiple regions, they can use Terraform's take on loops (count) to automatically create the same objects in multiple regions—is that the kind of thing you meant?

Something like that, but I could still see it getting messy with the `count`/`for_each` syntax. I think Bobby's solution:

> I think that in most cases, people only manage resources in 1 region, and if they wanted to replicate their whole setup in another region they would just change the default region provider configuration value.

Was what I was thinking of!

PR #618, opened 2024-07-27
Allow users to use cluster name as id

bobbyiliev:
This update introduces a new use_name_as_id option for the materialize_cluster resource. When set to true, this option allows the cluster name to be used as the Terraform resource ID instead of the internal cluster ID, e.g.:

```hcl
resource "materialize_cluster" "test_name_as_id" {
	name                = "test_name_as_id"
	size                = "25cc"
	replication_factor  = "1"
	identify_by_name    = true
}
```

The import functionality has also been updated to handle both name-based and ID-based imports, automatically setting `use_name_as_id` appropriately based on the import method used:

```sh
# Import using the name of the cluster so that the `identify_by_name ` is set to true automatically:
terraform import materialize_cluster.test_name_as_id aws/us-east-1:name:test_name_as_id
```

### DIFF
--- a/docs/resources/cluster.md
+++ b/docs/resources/cluster.md
@@ -64,8 +64,11 @@ Optional:
 Import is supported using the following syntax:
 
 ```shell
-# Clusters can be imported using the cluster id:
-terraform import materialize_cluster.example_cluster <region>:<cluster_id>
+# Clusters can be imported using the cluster id or name
+terraform import materialize_cluster.example_cluster <region>:id:<cluster_id>
+
+# To import using the cluster name, you need to set the `identify_by_name` attribute to true
+terraform import materialize_cluster.example_cluster <region>:name:<cluster_name>
 
 # Cluster id and information be found in the `mz_catalog.mz_clusters` table
 # The region is the region where the database is located (e.g. aws/us-east-1)
--- a/docs/resources/cluster.md
+++ b/docs/resources/cluster.md
@@ -30,6 +30,7 @@ resource "materialize_cluster" "example_cluster" {
 - `availability_zones` (List of String) The specific availability zones of the cluster.
 - `comment` (String) **Public Preview** Comment on an object in the database.
 - `disk` (Boolean, Deprecated) **Deprecated**. This attribute is maintained for backward compatibility with existing configurations. New users should use 'cc' sizes for disk access.
+- `identify_by_name` (Boolean) Use the cluster name as the Terraform resource ID instead of the internal cluster ID.
 - `introspection_debugging` (Boolean) Whether to introspect the gathering of the introspection data.
 - `introspection_interval` (String) The interval at which to collect introspection data.
 - `ownership_role` (String) The owernship role of the object.
@@ -37,7 +38,6 @@ resource "materialize_cluster" "example_cluster" {
 - `replication_factor` (Number) The number of replicas of each dataflow-powered object to maintain.
 - `scheduling` (Block List, Max: 1) Defines the scheduling parameters for the cluster. (see [below for nested schema](#nestedblock--scheduling))
 - `size` (String) The size of the managed cluster.
-- `use_name_as_id` (Boolean) Use the cluster name as the Terraform resource ID instead of the internal cluster ID.
 
 ### Read-Only
 
--- a/docs/resources/cluster.md
+++ b/docs/resources/cluster.md
@@ -37,6 +37,7 @@ resource "materialize_cluster" "example_cluster" {
 - `replication_factor` (Number) The number of replicas of each dataflow-powered object to maintain.
 - `scheduling` (Block List, Max: 1) Defines the scheduling parameters for the cluster. (see [below for nested schema](#nestedblock--scheduling))
 - `size` (String) The size of the managed cluster.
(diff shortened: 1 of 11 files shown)


docs/resources/cluster.md
```diff
@@ -37,6 +37,7 @@ resource "materialize_cluster" "example_cluster" {
 - `replication_factor` (Number) The number of replicas of each dataflow-powered object to maintain.
 - `scheduling` (Block List, Max: 1) Defines the scheduling parameters for the cluster. (see [below for nested schema](#nestedblock--scheduling))
 - `size` (String) The size of the managed cluster.
+- `use_name_as_id` (Boolean) Use the cluster name as the Terraform resource ID instead of the internal cluster ID.
```
benesch, line 40:
```suggestion
- `identify_by_name` (Boolean) Use the cluster name as the Terraform resource ID instead of the internal cluster ID.
```

On reflection, I propose we call the option `identify_by_name`—that just reads a little more smoothly to me than `use_name_as_id`!

pkg/resources/resource_cluster.go
```diff
@@ -339,3 +360,43 @@ func clusterDelete(ctx context.Context, d *schema.ResourceData, meta interface{}
...
+		// If not found by ID, try by name
+		s, err = materialize.ScanCluster(metaDb, importId, true)
+		if err != nil {
+			return nil, fmt.Errorf("error importing cluster %s: %s", importId, err)
+		}
+		// If found by name, set use_name_as_id to true
+		d.Set("use_name_as_id", true)
+	} else if err != nil {
+		return nil, fmt.Errorf("error importing cluster %s: %s", importId, err)
+	}
```
benesch, line 385: This looks a bit dangerous! What if someone names a cluster `u7`? This logic will get confused.

What I would recommend is baking the type of ID into the id string itself. Propose we add a prefix to the ID like `ID:` or `NAME:`

```
ID:u7
NAME:name
```

A resource ID that starts with `ID:` is a cluster ID, while a resource ID that starts with `NAME:` is a cluster name.

Then, when someone runs `terraform import <resource ID>`, you can look at the resource ID to determine whether you should look up by name or ID.

This would require a state migration, but I think a pretty straightforward one?

pkg/resources/resource_cluster.go
```diff
@@ -74,6 +75,13 @@ var clusterSchema = map[string]*schema.Schema{
 			},
 		},
 	},
+	"use_name_as_id": {
+		Type:        schema.TypeBool,
+		Optional:    true,
+		Default:     false,
+		Description: "Use the cluster name as the Terraform resource ID instead of the internal cluster ID.",
+		ForceNew:    true,
```
benesch, line 83: You probably _could_ handle in-place changes to this parameter, no? Just entails updating the resource ID.

pkg/provider/acceptance_cluster_test.go
```diff
@@ -447,7 +493,13 @@ func testAccCheckClusterExists(name string) resource.TestCheckFunc {
 		if !ok {
 			return fmt.Errorf("cluster not found: %s", name)
 		}
-		_, err = materialize.ScanCluster(db, utils.ExtractId(r.Primary.ID))
+		useNameAsId := false
+		if r.Primary.Attributes["use_name_as_id"] == "true" {
+			useNameAsId = true
+		}
+		// log the useNameAsId value
+		fmt.Printf("useNameAsId value: %v\n", useNameAsId)
```
benesch, line 501: Debugging?

benesch (commented):
Very nifty! Left some feedback in line.

pkg/resources/resource_cluster.go
```diff
@@ -87,29 +94,37 @@ func Cluster() *schema.Resource {
...
 	} else if err != nil {
 		return diag.FromErr(err)
 	}
 
-	d.SetId(utils.TransformIdWithRegion(string(region), i))
+	if useNameAsId {
+		d.SetId(utils.TransformIdWithTypeAndRegion(string(region), "name", s.ClusterName.String))
+	} else {
+		d.SetId(utils.TransformIdWithTypeAndRegion(string(region), "id", s.ClusterId.String))
+	}
```
bobbyiliev, line 127: There should be no need for a state migration, as we will adjust the ID format accordingly directly in the read func.

docs/resources/cluster.md
```diff
@@ -63,8 +64,11 @@ Optional:
 Import is supported using the following syntax:
 
 ```shell
-# Clusters can be imported using the cluster id:
-terraform import materialize_cluster.example_cluster <region>:<cluster_id>
+# Clusters can be imported using the cluster id or name
+terraform import materialize_cluster.example_cluster <region>:id:<cluster_id>
+
+# To import using the cluster name, you need to set the `identify_by_name` attribute to true
+terraform import materialize_cluster.example_cluster <region>:name:<cluster_name>
```
benesch, line 71: This looks so clean!

pkg/resources/resource_cluster.go
benesch: Ah ok, nice!

benesch (approved):
Gorgeous, Bobby, thank you!

PR #626, opened 2024-08-06
Fix duplicate application name

### LINKED ISSUE #625: Bug: the provider seems to send a duplicate application name
The Terraform provider seems to send a duplicate application name:

```
terraform-provider-materialize v0.8.5 terraform-provider-materialize v0.8.5
```

The problem seems to be in the `NewDBClient` method:

https://github.com/MaterializeInc/terraform-provider-materialize/blob/a74af7470cdc219191d2edf749df511293484c43/pkg/provider/provider.go#L217

https://github.com/MaterializeInc/terraform-provider-materialize/blob/a74af7470cdc219191d2edf749df511293484c43/pkg/clients/db_client.go#L18

bobbyiliev:
Fixes #625

### DIFF
--- a/pkg/clients/db_client.go
+++ b/pkg/clients/db_client.go
@@ -12,13 +12,10 @@ type DBClient struct {
 	*sqlx.DB
 }
 
-func NewDBClient(host, user, password string, port int, database, application_name_suffix, version, sslmode string) (*DBClient, diag.Diagnostics) {
+func NewDBClient(host, user, password string, port int, database, application_name, version, sslmode string) (*DBClient, diag.Diagnostics) {
 	var diags diag.Diagnostics
 
-	var application_name string
-	if application_name_suffix != "" {
-		application_name = application_name_suffix
-	} else {
+	if application_name == "" {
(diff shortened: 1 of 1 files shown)


pkg/clients/db_client.go
```diff
@@ -15,9 +15,11 @@ type DBClient struct {
 func NewDBClient(host, user, password string, port int, database, application_name_suffix, version, sslmode string) (*DBClient, diag.Diagnostics) {
 	var diags diag.Diagnostics
 
-	application_name := fmt.Sprintf("terraform-provider-materialize v%s", version)
+	var application_name string
 	if application_name_suffix != "" {
-		application_name += " " + application_name_suffix
+		application_name = application_name_suffix
```
RobinClowers, line 20: It seems confusing to call this _suffix if it's actually the full application name?

RobinClowers (approved):

pkg/clients/db_client.go
bobbyiliev: Good point! I changed this a bit as it indeed did not make sense to have it this way

PR #628, opened 2024-08-08
unify cluster alter commands

jubrad:
Unifying cluster alter commands. This is a pre-req for adding graceful reconfiguration.

### DIFF
--- a/pkg/materialize/cluster.go
+++ b/pkg/materialize/cluster.go
@@ -31,6 +31,7 @@ func NewClusterBuilder(conn *sqlx.DB, obj MaterializeObject) *ClusterBuilder {
 
 type SchedulingConfig struct {
 	OnRefresh OnRefreshConfig
+	IsSet     bool
 }
 
 type OnRefreshConfig struct {
@@ -68,6 +69,7 @@ func GetSchedulingConfig(v interface{}) SchedulingConfig {
 		OnRefresh: OnRefreshConfig{
 			Enabled: false,
 		},
+		IsSet: true,
 	}
 
 	if enabled, ok := onRefreshMap["enabled"].(bool); ok {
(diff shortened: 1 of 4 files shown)


jubrad:
Test this with the following setup:
# ./terraformrc
```
provider_installation {
  dev_overrides {
    "MaterializeInc/materialize" = "/Users/justin/go/bin/"
  }
  direct {}
}
```

```hcl
terraform {
  required_providers {
    materialize = {
      source = "MaterializeInc/materialize"
    }
  }
}

provider "materialize" {
  password       = "dontlookatmypassworddood"
  default_region = "aws/us-east-1"
  database       = "materialize"
  cloud_endpoint = "https://api.staging.cloud.materialize.com"
  base_endpoint  = "https://admin.staging.cloud.materialize.com"
  endpoint       = "https://admin.staging.cloud.materialize.com"
}

resource "materialize_cluster" "test" {
  name               = "test"
  size               = "25cc"
  replication_factor = 2
}
```
Then changed the size/replication factor a few ways..
I can see in the sql lifecycle on the console that it ran successfully in a single statement, where using the latest release was multiple statements.
image

pkg/materialize/cluster.go
```diff
@@ -191,57 +210,32 @@ func (b *ClusterBuilder) Drop() error {
...
-		if schedulingConfig.OnRefresh.HydrationTimeEstimate != "" {
-			scheduleStatement += fmt.Sprintf(" (HYDRATION TIME ESTIMATE = %s)", QuoteString(schedulingConfig.OnRefresh.HydrationTimeEstimate))
-		} else if schedulingConfig.OnRefresh.RehydrationTimeEstimate != "" {
-			scheduleStatement += fmt.Sprintf(" (HYDRATION TIME ESTIMATE = %s)", QuoteString(schedulingConfig.OnRefresh.RehydrationTimeEstimate))
-		}
-		q = fmt.Sprintf("ALTER CLUSTER %s SET (%s);", b.QualifiedName(), scheduleStatement)
-	} else {
-		// Reset the schedule settings if not enabled.
-		q = fmt.Sprintf("ALTER CLUSTER %s RESET (SCHEDULE);", b.QualifiedName())
-	}
```
bobbyiliev, line 242: This looks great! Thanks a lot @jubrad!

The only issue that I've noticed while manually testing this is that if you were to change the scheduling from enabled true to false, it generates the following alter statement which will fail:

```sql
ALTER CLUSTER "scheduling_cluster" SET ();
```

Repro:

```hcl
resource "materialize_cluster" "scheduling_cluster" {
  name = "scheduling_cluster"
  size = "50cc"
  scheduling {
    on_refresh {
      enabled                 = true # Apply and then change to false and apply again
      hydration_time_estimate = "3 hour"
    }
  }
}
```

pkg/materialize/cluster.go
```diff
@@ -124,65 +126,88 @@ func (b *ClusterBuilder) Scheduling(v []interface{}) *ClusterBuilder {
...
+	return b.ddl.exec(q.String())
+}
+
+func (b *ClusterBuilder) AlterClusterScheduling(s SchedulingConfig) error {
+	q := strings.Builder{}
+	q.WriteString(fmt.Sprintf(`ALTER CLUSTER %s`, b.QualifiedName()))
+	q.WriteString(fmt.Sprintf(` SET (%s)`, b.GenSchedulingConfigSql(s)))
+	q.WriteString(`;`)
+	return b.ddl.exec(q.String())
+}
```
jubrad, line 200: @bobbyiliev the plan now is that we do AlterClusterScheduling separate from AlterCluster and always first. 

This will allow us to from a cluster with scheduling and no replication factor to a cluster without scheduling and a replication factor, and vice versa.

Note, we only really set the scheduling option on the ClusterBuilder directly in the create phase where everything can be done in a single statement.

bobbyiliev (approved):
I believe that this looks good! Thank you for putting this all together @jubrad 👏

PR #630, opened 2024-08-09
Fix AWS connection empty internal table

### LINKED ISSUE #629: Bug: `materialize_connection_aws` resource read query not detecting all attributes
After creating an AWS connection using the `materialize_connection_aws` resource, there are no entries in the `mz_internal.mz_aws_connections`table.

https://github.com/MaterializeInc/terraform-provider-materialize/blob/183b8d8d5f1ee06d9e5c64e03a0725ae6aff2e05/pkg/materialize/connection_aws.go#L143-L175

We should investigate why this is happening and if the view name has changed or if this is an actual issue.

bobbyiliev:
Fixes #629

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -17,6 +17,7 @@ services:
       - --availability-zone=test2
       - --aws-account-id=123456789000
       - --aws-external-id-prefix=eb5cb59b-e2fe-41f3-87ca-d2176a495345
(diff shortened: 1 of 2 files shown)


benesch (approved):
Ah yeah sorry, I forgot about the other --aws arg you had to pass. LGTM!

PR #632, opened 2024-08-14
Feature/graceful reconfig

jubrad:
Adds `wait_until_ready` option to cluster resources, which will help reduce, ideally eliminating, downtime when terraform updates a cluster; for example, changes the size. The default will be to turn this off.

ex:
```hcl
resource "materialize_cluster" "cluster" {
  name = var.mz_cluster
  size = "25cc"
  wait_until_ready {
    enabled = true
    timeout = "10m"
    on_timeout = "COMMIT"
  }
}

```

This will be gated on a feature flag, and is not yet ready for primetime, we still have a few [PRs](https://github.com/MaterializeInc/materialize/pull/28836) that need to go in on the database side before this functionality is available. This  pr is also stacked on https://github.com/MaterializeInc/terraform-provider-materialize/pull/628, and will need to be rebased once that goes in.


bobbyiliev (commented):
This looks very good! 👏 

We should probably add an acceptance test to get some better coverage.

bobbyiliev (approved):
Thanks again for this PR @jubrad!

PR #636, opened 2024-08-26
Rename docker-compose.yml to compose.yaml

bobbyiliev:
As per the new Docker naming convention described in the docs [here](https://docs.docker.com/compose/compose-application-model/#the-compose-file), renaming the `docker-compose.yml` to `compose.yaml`.

> The default path for a Compose file is `compose.yaml` (preferred) or compose.yml that is placed in the working directory. Compose also supports `docker-compose.yaml` and `docker-compose.yml` for backwards compatibility of earlier versions. If both files exist, Compose prefers the canonical `compose.yaml`.


arusahni (approved):
How long has support for `compose.yml` been out in the wild? I wonder if this could break things for people using older tooling that doesn't yet support the new convention.

bobbyiliev:
> How long has support for `compose.yml` been out in the wild? I wonder if this could break things for people using older tooling that doesn't yet support the new convention.

Yea, good point! Managed to find when the support for the `compose.yml` and `compose.yaml` was actually introduced and it seems like it's been there for quite a while already: https://docs.docker.com/compose/release-notes/#1286

My guess is that they will keep the `docker-compose.yaml` fallback for quite a while!

PR #638, opened 2024-08-28
Fix user not found state status

### LINKED ISSUE #637: `materialize_user` resource fails with 404 error if manually deleted
When a user is created using the `materialize_user` resource in Terraform, and that user is later deleted manually outside of Terraform, the Terraform state becomes stuck. Subsequent Terraform operations fail with the following error:

```
╷
│ Error: HTTP request error: status 404, response: 
│ 
│   with materialize_user.example_user,
│   on main.tf in resource "materialize_user" "example_user":
│  167: resource "materialize_user" "example_user" {
```

**Steps to Reproduce:**
1. Create a user with the `materialize_user` Terraform resource.
2. Manually delete the user using the Frontegg API or the console
3. Run any Terraform operation that interacts with the `materialize_user` resource.

**Expected Behavior:**
Terraform should handle the 404 error gracefully, either by removing the resource from the state or by providing a clear path for the user to resolve the state.

**Actual Behavior:**
Terraform state gets stuck, and further operations fail with a 404 error.

bobbyiliev:
Fixes #637 

```
│ Error: HTTP request error: status 404, response: 
│ 
│   with materialize_user.example_user,
│   on main.tf in resource "materialize_user" "example_user":
│  167: resource "materialize_user" "example_user" {
```

### DIFF
--- a/pkg/resources/resource_user.go
+++ b/pkg/resources/resource_user.go
@@ -123,7 +123,8 @@ func userRead(ctx context.Context, d *schema.ResourceData, meta interface{}) dia
 
 	userResponse, err := frontegg.ReadUser(ctx, client, userID)
 	if err != nil {
-		if strings.Contains(err.Error(), "not found") {
+		if strings.Contains(err.Error(), "404") || strings.Contains(strings.ToLower(err.Error()), "not found") {
+			// User doesn't exist, remove from state
 			d.SetId("")
 			return nil
 		}


pkg/resources/resource_user.go
```diff
@@ -123,7 +123,8 @@ func userRead(ctx context.Context, d *schema.ResourceData, meta interface{}) dia
 
 	userResponse, err := frontegg.ReadUser(ctx, client, userID)
 	if err != nil {
-		if strings.Contains(err.Error(), "not found") {
+		if strings.Contains(err.Error(), "404") || strings.Contains(strings.ToLower(err.Error()), "not found") {
+			// User doesn't exist, remove from state
```
arusahni, line 127: Digging into the Frontegg helpers, it seems as if we juggling this a few ways:

1. `ReadUser` calls `doRequest`, [which returns](https://github.com/MaterializeInc/terraform-provider-materialize/blob/main/pkg/frontegg/helpers.go#L31-L39) `nil, fmt.Errorf(...)` for 404 errors.
2. Since `err != nil`, [we immediately return](https://github.com/MaterializeInc/terraform-provider-materialize/blob/08b526afd8a7c0c44e07953ecd6d6d071220ec34/pkg/frontegg/user.go#L90-L92) that same result from `ReadUser`
3. Further along in `ReadUser` (after we've early-returned out), [we invoke an API error handler](https://github.com/MaterializeInc/terraform-provider-materialize/blob/08b526afd8a7c0c44e07953ecd6d6d071220ec34/pkg/frontegg/user.go#L95-L97) for non-200 status codes and return a string with different formatting from the Frontegg helper in step 1.

Ideally we can consolidate this string formatting to be consistent across endpoints, or even include a status code (by using an Error struct).

arusahni (approved):
My feedback is more of a nice-to-have than something that should be a blocker in getting this bugfix out.

pkg/resources/resource_user.go
bobbyiliev: Good catch @arusahni! Thank you for this!

I've created a tracking issue for this: https://github.com/MaterializeInc/terraform-provider-materialize/issues/641

PR #640, opened 2024-08-28
Disable telemetry in CI

bobbyiliev:
Disabling telemetry in CI:

https://github.com/MaterializeInc/materialize/blob/bf068578fb2fa309621039c0529716efbe8bfa08/misc/images/materialized/entrypoint.sh#L82


benesch (approved):
Thanks!

PR #644, opened 2024-09-02
 resource: add password to user create route

### LINKED ISSUE #643: `materialize_user`: introduce a `password` attribute
### Feature request

There isn't currently a way to create a new user with a predefined password, however this is supported by the Frontegg API:

> https://docs.frontegg.com/reference/userscontrollerv2_createuser

We should extend the `materialize_user` resource and introduce an optional `password` attribute.

bobbyiliev:
Closes #643 

As per the discussion [here](https://github.com/MaterializeInc/materialize/pull/29331), I now remember why I did not initially implement this on the Terraform side, if you were to try and create a user with a pre-defined password using the staging Frontegg API you would get:

```
Failed to create user: Frontegg API error (HTTP 400): unexpected status code: 400 - {"errors":["User creation with password is forbidden"]}
```

We could still have this supported in the mock service, but it will diverge from the staging/prod Frontegg API.

### DIFF
--- a/docs/resources/user.md
+++ b/docs/resources/user.md
@@ -29,6 +29,7 @@ resource "materialize_user" "example_user" {
 
 ### Optional
 
+- `password` (String, Sensitive) The password for the user. If not provided, an activation email will be sent.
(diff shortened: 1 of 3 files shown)


bobbyiliev:
Closing this PR as we've agreed not to diverge from the Frontegg API

PR #645, opened 2024-09-02
Add source versioning design doc

bobbyiliev:
As per the  Source Versioning / Tables from Sources changes: https://github.com/MaterializeInc/materialize/pull/27907

### DIFF
--- a/docs/developer/source-versioning-2024-08-30.md
+++ b/docs/developer/source-versioning-2024-08-30.md
@@ -37,7 +37,7 @@ var sourcePostgresSchema = map[string]*schema.Schema{
         Description: "Tables to be ingested from the source. This field is deprecated and will be removed in a future version.",
         Type:        schema.TypeSet,
         Optional:    true,
-        Deprecated:  "Use the new `materialize_table_from_source` resource instead.",
+        Deprecated:  "Use the new `materialize_source_table` resource instead.",
         Elem: &schema.Resource{
             Schema: map[string]*schema.Schema{
                 // ... existing table schema ...
@@ -49,10 +49,10 @@ var sourcePostgresSchema = map[string]*schema.Schema{
 
 #### 4.1.2 New Table From Source Resource
 
-Introduce a new `materialize_table_from_source` resource:
+Introduce a new `materialize_source_table` resource:
 
 ```go
-var tableFromSourceSchema = map[string]*schema.Schema{
+var sourceTableSchema = map[string]*schema.Schema{
     "name":               ObjectNameSchema("table", true, false),
     "schema_name":        SchemaNameSchema("table", false),
     "database_name":      DatabaseNameSchema("table", false),
@@ -87,21 +87,21 @@ var tableFromSourceSchema = map[string]*schema.Schema{
 
 ### 4.2 Resource Implementation
 
-#### 4.2.1 Table From Source Resource
+#### 4.2.1 Table Source Table Resource
 
-Implement CRUD operations for the new `materialize_table_from_source` resource:
+Implement CRUD operations for the new `materialize_source_table` resource:
 
 ```go
-func TableFromSource() *schema.Resource {
+func SourceTable() *schema.Resource {
     return &schema.Resource{
-        CreateContext: tableFromSourceCreate,
-        ReadContext:   tableFromSourceRead,
-        UpdateContext: tableFromSourceUpdate,
-        DeleteContext: tableFromSourceDelete,
+        CreateContext: sourceTableCreate,
+        ReadContext:   sourceTableRead,
+        UpdateContext: sourceTableUpdate,
+        DeleteContext: sourceTableDelete,
         Importer: &schema.ResourceImporter{
             StateContext: schema.ImportStatePassthroughContext,
         },
-        Schema: tableFromSourceSchema,
+        Schema: sourceTableSchema,
     }
 }
 ```
@@ -116,7 +116,7 @@ CREATE TABLE <database_name>.<schema_name>.<name> FROM SOURCE <source_name> (REF
 
 Modify the CRUD operations for existing source resources to handle the deprecation of subsource-related fields:
 
-- In `Create` and `Update` operations, if the deprecated `table` field is used, log a warning message advising users to migrate to the new `materialize_table_from_source` resource.
+- In `Create` and `Update` operations, if the deprecated `table` field is used, log a warning message advising users to migrate to the new `materialize_source_table` resource.
 - In `Read` operations, continue to populate the `table` field if it exists in the state, but also log a deprecation warning.
 
 ### 4.3 Migration Strategy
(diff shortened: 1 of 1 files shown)


docs/developer/source-versioning-2024-08-30.md
```diff
@@ -0,0 +1,183 @@
...
+                // ... existing table schema ...
+            },
+        },
+    },
+}
+```
+
+#### 4.1.2 New Table From Source Resource
+
+Introduce a new `materialize_table_from_source` resource:
```
benesch, line 52: I would call it `materialize_source_table`; the "table from source" is much more SQL-y than it is Terraform-y.

docs/developer/source-versioning-2024-08-30.md
```diff
@@ -0,0 +1,183 @@
...
+
+1. Deprecate the `table` field in existing source resources.
+2. Introduce the new `materialize_table_from_source` resource.
+3. Allow both old and new configurations to coexist during a transition period.
+
+This approach allows users to migrate their configurations gradually:
+
+- Existing sources can still be created and managed.
+- New tables (formerly subsources) will be created as separate `materialize_table_from_source` resources.
+- Users can migrate their configurations at their own pace by replacing `table` blocks with `materialize_table_from_source` resources.
```
benesch, line 134: This is really smart! Will you be able to mix-n-match `table` and `materialize_source_table` resources for the same source? Or will you always have to migrate an entire source at once?

My hunch is that we should require that you only use one or the other on a per source basis. That seems likely to make the code much simpler and easier to reason about. And it still affords the user quite a bit of flexibility to do the migration incrementally by migrating one source at a time.

benesch (commented):
This all makes sense to me! Only took a quick skim though. Deferring the full review to the experts (@rjobanp and @morsapaes).

docs/developer/source-versioning-2024-08-30.md
bobbyiliev: Oh nice, I like this way better! Thanks for that suggestion!

docs/developer/source-versioning-2024-08-30.md
bobbyiliev: >  Will you be able to mix-n-match table and materialize_source_table resources for the same source? Or will you always have to migrate an entire source at once?

Based on what I've tested so far using main, it looks like that you could have both running simultaneously. So if we were to take the following example:

- Existing standard Postgres source:

```hcl
resource "materialize_source_postgres" "pg_source" {
  name         = "pg_source"
  cluster_name = "quickstart"

  expose_progress {
    name = "expose_postgres_progress"
  }

  postgres_connection {
    name          = materialize_connection_postgres.postgres_connection.name
  }
  publication = "mz_source"
  table {
    upstream_name  = "table1"
    upstream_schema_name = "public"
  }
  table {
    upstream_name  = "table2"
    upstream_schema_name = "public"
  }
  ...
}
```

- Then define a new `materialize_source_table`:

```hcl
resource "materialize_source_table" "source_table3" {
  name           = "source_table3"
  schema_name    = "public"
  database_name  = "materialize"

  source {
    name          = materialize_source_postgres.pg_source.name
    schema_name   = "public"
    database_name = "materialize"
  }

  upstream_name         = "table2"
  upstream_schema_name  = "public"

  text_columns = [
    "updated_at"
  ]
}
```

So the user should still be able to slowly do the migration if needed as the `table` blocks and the `materialize_source_table` are independent, at least until we do the actual migration on the Materialize side, then in that case, users would just need to define all of their tables as `materialize_source_table` and as long as we have all of the required information in the catalog tables, we should be able to update the state accordingly.

rjobanp: >  it looks like that you could have both running simultaneously

Yes at the moment we allow both if you have the new `enable_create_table_from_source` feature-flag enabled, however we could artificially enforce that you are using either the old model or the new model in the terraform provider on a per-source basis to make things simpler?

docs/developer/source-versioning-2024-08-30.md
```diff
@@ -0,0 +1,183 @@
...
+
+1. Deprecate the `table` field in existing source resources.
+2. Introduce the new `materialize_source_table` resource.
+3. Allow both old and new configurations to coexist during a transition period.
+
+This approach allows users to migrate their configurations gradually:
+
+- Existing sources can still be created and managed.
+- New tables (formerly subsources) will be created as separate `materialize_source_table` resources.
+- Users can migrate their configurations at their own pace by replacing `table` blocks with `materialize_source_table` resources.
```
rjobanp, line 134: this makes sense to me to for allowing users to opt-in to the migration on their own accord!

We've also discussed implementing a catalog migration on the MZ side to automatically migrate all existing sources/subsources over to the new model, and in that case we would alter the `CREATE SOURCE` statement to remove the deprecated `table` field options. If the user reruns terraform after that migration it would attempt to re-introduce the `table` field unless the user was proactive in updating their terraform config in coordination with that migration. Do we think there'd be a way to detect this situation? Perhaps if the `table` field is already empty and the `enable_create_table_from_source ` feature-flag is enabled we just ignore the terraform config for that option and emit a warning?

docs/developer/source-versioning-2024-08-30.md
morsapaes: 💯, that's also what we'll use for the dbt materialization (`source-table`).

docs/developer/source-versioning-2024-08-30.md
benesch: Yeah, in general these migrations are so hairy when they involve stuff happening inside Materialize and stuff happening with Terraform. 

I don't know that there will be much value in teaching the Terraform configuration to ignore the `table` field once the migration occurs. There's no guarantee that users will upgrade to a version of the Terraform provider that has the special handling for `table` before we do the catalog migration. And the catalog migration will happen behind Terraform's back, which is pretty sketchy—someone who thought they had everything properly tracked in Terraform will suddenly find that is not the case.

I think our best bet here may be to support the two syntaxes indefinitely, but disable the old syntax for _new_ sources as soon as we can pull off. That will prevent the problem from getting any worse. Then, after a couple months, hopefully there are few enough sources remaining that use the old syntax that we can go customer by customer and source by source and figure out if they're using Terraform to manage the source or not. If they're using Terraform, I think we're better off having field eng work with them to get through the upgrade manually than trying to do anything automatic.

docs/developer/source-versioning-2024-08-30.md
benesch: Ah ok, if it's going to be supported to have both simultaneously on the database side, then maybe it's actually simpler to exactly mirror that in Terraform?

bobbyiliev: Yep, it would be simpler if we could have it as it is at the moment. Allowing us to have both simultaneously on the database side makes things easier for the Terraform implementation as well.

docs/developer/source-versioning-2024-08-30.md
bobbyiliev: > I think our best bet here may be to support the two syntaxes indefinitely, but disable the old syntax for new sources as soon as we can pull off.

This sounds very good to me!

>  I think we're better off having field eng work with them to get through the upgrade manually than trying to do anything automatic.

I will try to prepare a step by step migrations guide on how to switch from the old to the new syntax on the Terraform side.

morsapaes (approved):

Issue #646, opened 2024-09-02
Add Support for Source Versioning / Tables from Sources


bobbyiliev:
### Feature request

As per the discussion here: https://github.com/MaterializeInc/materialize/pull/27907 and based on the design doc here: https://github.com/MaterializeInc/terraform-provider-materialize/pull/645 we need to introduce the support for tables from sources on the Terraform provider side as well.

Extra details to be discussed in the design doc pull request, but the main things to keep in mind are:

- [ ] Deprecated the related table attributes on the source resources, eg. MySQL, Postgres, Loadgens and Kafka
- [ ] Introduce the new `resource_table_from_source` resource
- [ ] Add tests including unit tests, integration tests and acceptance tests
- [ ] Update all of the related documentation
- [ ] Write a migrations guide for users using the old implementation

Issue #391, opened 2023-11-30
Postgres source: consider renaming `schema` and `table` options


morsapaes:
The `schema` option takes a list of strings, so it'd be less confusing to call it `schemas` (= `FOR SCHEMAS`). The same is true for `table` (= `FOR TABLES`). 

Not sure if there's a way to change this in a backwards-compatible way. It might be worth double-checking if there are similar examples and patch them all in one go.

dehume:
No good way to make this backward-compatible. The best I can think of is include the plural as well and include a deprecation warning for the singular (will look at how other providers handled this). Will also compile the list of singular vs plural attribute names. We do have situations where we use the singular for an attribute but allow multiple blocks, but that is usually just for nested elements like columns in a table.
```
resource "materialize_table" "simple_table" {
  name          = "simple_table"
  column {
    name = "column_1"
    type = "text"
  }
  column {
    name    = "column_2"
    type    = "int"
  }
```

dehume:
**Lists** For attributes that use list, we have more cases of singular than plural.
| Attribute | Resource | Type | Plural |
| --- | --- | --- | --- |
| availability_zones | materialize_connection_aws_privatelink | List of Strings | X |
| not_null_assertions | materialize_materialized_view | List of Strings | X |
| start_offset | materialize_source_kafka | List of Strings | |
| header | materialize_source_kafka | List of Strings | |
| subsource | materialize_source_kafka | List of Objects | |
| subsource | materialize_source_load_generator | List of Objects | |
| schema | materialize_source_postgres | List of Strings | |
| text_columns | materialize_source_postgres | List of Strings | X |
| subsource | materialize_source_postgres | List of Objects | |
| subsource | materialize_webhook | List of Objects | |

**Blocks** We had decided should be singular. There are some blocks that use plural 
| Attribute | Resource | Type | Plural |
| --- | --- | --- | --- |
| auction_options | materialize_source_load_generator | Block | X |
| counter_options | materialize_source_load_generator | Block | X |
| marketing_options | materialize_source_load_generator | Block | X |
| tpch_options | materialize_source_load_generator | Block | X |
| check_options | materialize_webhook | Block | X |
| row_properties | materialize_type | Block | X |
| map_properties | materialize_type | Block | X |
| list_properties | materialize_type | Block | X |

dehume:
[...]

PR #647, opened 2024-09-03
Source versioning: Postgres, MySQL and Load generator

bobbyiliev:
Initial implementation for the source versioning refactor as per #646 

The main changes to consider:
- Marking the `table` attribute as optional and deprecated for both the MySQL and Postgres sources
- Introduced a new `all_tables` bool attribute for the MySQL and the Loadgen sources, as in the past this was defaulting always using `FOR ALL TABLES` in the load gen sources (auction, marketing, tpch) and in the MySQL case whenever no `table` blocks were defined, we defaulted to `FOR ALL TABLES`. This `all_tables` bool attribute allows us to create sources without any tables defined as per the source versioning work
- Introducing the new `materialize_source_table_{mysql|postgres|load_generator}` resource which allows us to do `CREATE TABLE ... FROM SOURCE ...`

Things that are still pending: https://github.com/MaterializeInc/terraform-provider-materialize/issues/646


docs/resources/source_table.md
```diff
@@ -29,7 +29,7 @@ description: |-
 - `ownership_role` (String) The owernship role of the object.
 - `region` (String) The region to use for the resource connection. If not set, the default region is used.
 - `schema_name` (String) The identifier for the table schema in Materialize. Defaults to `public`.
-- `text_columns` (List of String) Columns to be decoded as text.
+- `text_columns` (List of String) Columns to be decoded as text. Not supported for the load generator sources, if the source is a load generator, the attribute will be ignored.
```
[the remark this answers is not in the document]
bobbyiliev: Yes indeed, I was just thinking about this. With the MySQL and Postgres sources, it is probably fine, but as soon as we add Kafka and Webhook sources, the logic will get out of hand.

Will refactor this to have a separate source table resource per source!

docs/resources/source_kafka.md
```diff
@@ -56,25 +56,25 @@ resource "materialize_source_kafka" "example_source_kafka" {
...
+- `include_timestamp_alias` (String, Deprecated) Provide an alias for the timestamp column. Deprecated: Use the new materialize_source_table_kafka resource instead.
 - `key_format` (Block List, Max: 1) Set the key format explicitly. (see [below for nested schema](#nestedblock--key_format))
 - `ownership_role` (String) The owernship role of the object.
 - `region` (String) The region to use for the resource connection. If not set, the default region is used.
 - `schema_name` (String) The identifier for the source schema in Materialize. Defaults to `public`.
-- `start_offset` (List of Number) Read partitions from the specified offset.
-- `start_timestamp` (Number) Use the specified value to set `START OFFSET` based on the Kafka timestamp.
+- `start_offset` (List of Number, Deprecated) Read partitions from the specified offset. Deprecated: Use the new materialize_source_table_kafka resource instead.
+- `start_timestamp` (Number, Deprecated) Use the specified value to set `START OFFSET` based on the Kafka timestamp. Deprecated: Use the new materialize_source_table_kafka resource instead.
```
rjobanp, line 77: these two options are still currently only possible on the top-level `CREATE SOURCE` statement for kafka sources -- not yet on a per-table basis. It will require a non-trivial amount more refactoring to allow them on a per-table basis so I'm unsure if we will do that work until it's requested by a customer

docs/resources/source_mysql.md
```diff
@@ -53,12 +53,12 @@ resource "materialize_source_mysql" "test" {
 - `comment` (String) **Public Preview** Comment on an object in the database.
 - `database_name` (String) The identifier for the source database in Materialize. Defaults to `MZ_DATABASE` environment variable if set or `materialize` if environment variable is not set.
 - `expose_progress` (Block List, Max: 1) The name of the progress collection for the source. If this is not specified, the collection will be named `<src_name>_progress`. (see [below for nested schema](#nestedblock--expose_progress))
-- `ignore_columns` (List of String, Deprecated) Ignore specific columns when reading data from MySQL. Can only be updated in place when also updating a corresponding `table` attribute. Deprecated: Use the new materialize_source_table resource instead.
+- `ignore_columns` (List of String, Deprecated) Ignore specific columns when reading data from MySQL. Can only be updated in place when also updating a corresponding `table` attribute. Deprecated: Use the new materialize_source_table_mysql resource instead.
```
rjobanp, line 56: fyi this option is also being renamed https://github.com/MaterializeInc/materialize/pull/29438 but the old name will be aliased to the new one, so this shouldn't break

docs/resources/source_table_kafka.md
```diff
@@ -0,0 +1,325 @@
...
+- `include_partition` (Boolean) Include a partition column containing the Kafka message partition
+- `include_partition_alias` (String) Provide an alias for the partition column.
+- `include_timestamp` (Boolean) Include a timestamp column containing the Kafka message timestamp.
+- `include_timestamp_alias` (String) Provide an alias for the timestamp column.
+- `key_format` (Block List, Max: 1) Set the key format explicitly. (see [below for nested schema](#nestedblock--key_format))
+- `ownership_role` (String) The owernship role of the object.
+- `region` (String) The region to use for the resource connection. If not set, the default region is used.
+- `schema_name` (String) The identifier for the source schema in Materialize. Defaults to `public`.
+- `start_offset` (List of Number) Read partitions from the specified offset.
+- `start_timestamp` (Number) Use the specified value to set `START OFFSET` based on the Kafka timestamp.
```
rjobanp, line 46: these aren't currently available on a per-table basis for kafka sources

docs/resources/source_table_kafka.md
```diff
@@ -0,0 +1,325 @@
...
+- `upstream_schema_name` (String) The schema of the table in the upstream database.
```
rjobanp, line 47: what does this refer to for kafka sources? we might just want to omit it since the upstream reference should just be the kafka topic name

pkg/materialize/source_table_kafka.go
```diff
@@ -0,0 +1,332 @@
...
+	headersAlias     string
+	partitionAlias   string
+	offsetAlias      string
+	timestampAlias   string
+	format           SourceFormatSpecStruct
+	keyFormat        SourceFormatSpecStruct
+	valueFormat      SourceFormatSpecStruct
+	envelope         KafkaSourceEnvelopeStruct
+	startOffset      []int
+	startTimestamp   int
```
rjobanp, line 27: these two aren't used below and also aren't possible on the statement

docs/guides/materialize_source_table.md
```diff
@@ -0,0 +1,195 @@
...
+
+---
+
+# Source versioning: migrating to `materialize_source_table_{source}` Resource
+
+In previous versions of the Materialize Terraform provider, source tables were defined within the source resource itself and were considered subsources of the source rather than separate entities.
+
+This guide will walk you through the process of migrating your existing source table definitions to the new `materialize_source_table_{source}` resource.
+
+For each source type (e.g., MySQL, Postgres, etc.), you will need to create a new `materialize_source_table_{source}` resource for each table that was previously defined within the source resource. This ensures that the tables are preserved during the migration process.
```
rjobanp, line 14:
```suggestion
For each source type (e.g., MySQL, Postgres, etc.), you will need to create a new `materialize_source_table_{source}` resource for each table that was previously defined within the source resource. This ensures that the tables are preserved during the migration process. For Kafka sources, you will need to create at least one `materialize_source_table_kafka` table to hold data for the kafka topic.
```
@morsapaes might have better wording for this but I think we should be clear that this migration needs to happen for sources that previously didn't have subsources too (e.g. kafka)

docs/guides/materialize_source_table.md
```diff
@@ -0,0 +1,195 @@
...
+
+### Step 5: Verify the Migration
+
+After applying the changes, verify that your tables are still correctly set up in Materialize by checking the table definitions using Materialize’s SQL commands.
+
+During the migration, you can use both the old `table` blocks and the new `materialize_source_table_{source}` resources simultaneously. This allows for a gradual transition until the old method is fully deprecated.
+
+The same approach can be used for other source types such as Postgres, eg. `materialize_source_table_postgres`.
+
+## Automated Migration Process (TBD)
```
rjobanp, line 116: nice - this is great! We will probably want to figure out how to tell them that they will be able to coordinate the 'automated' migration process with their field-engineer representative if they go down this path

rjobanp (commented):
nice work!

docs/resources/source_kafka.md
bobbyiliev: Ah yes! Good catch! Thank you!

docs/resources/source_mysql.md
bobbyiliev: Sounds good! I will go ahead and use the exclude columns for the new table source resource!

docs/resources/source_table_kafka.md
bobbyiliev: Good catch, this was an overlook on my end in the schema for the Kafka source table resource.

bobbyiliev:
One thing that we can consider here as per this old tracking issue: https://github.com/MaterializeInc/terraform-provider-materialize/issues/391 is take the chance and decide if we still want to rename some of the attributes in the new source table resources:

**Lists** For attributes that use list, we have more cases of singular than plural.
| Attribute | Resource | Type | Plural | Comment |
| --- | --- | --- | --- | --- |
| start_offset | materialize_source_kafka | List of Strings | | In Materialize the attribute is singular `START OFFSET` even though it is a list of strings |
| header | materialize_source_kafka | List of Strings | | In Materialize the attribute is singular `HEADER` even though it is a list of strings |
| text_columns | materialize_source_postgres | List of Strings | X | |

**Blocks** We had decided should be singular. There are some blocks that use plural so this could be a good chance to rename those attributes in the new source table load gen resource:
| Attribute | Resource | Type | Plural |
| --- | --- | --- | --- |
| auction_options | materialize_source_load_generator | Block | X |
| counter_options | materialize_source_load_generator | Block | X |
| marketing_options | materialize_source_load_generator | Block | X |
| tpch_options | materialize_source_load_generator | Block | X |
| check_options | materialize_webhook | Block | X |

docs/guides/materialize_source_table.md
arusahni, line 1: This appears to be a rendered version of `templates/guides/materialize_source_table.md.tmpl` but doesn't have the header that states it's generated. We should probably add that so no-one edits this file directly.

pkg/materialize/format_specs.go
```diff
@@ -73,7 +73,7 @@ func GetFormatSpecStruc(v interface{}) SourceFormatSpecStruct {
 	}
 	if protobuf, ok := u["protobuf"]; ok && protobuf != nil && len(protobuf.([]interface{})) > 0 {
 		if csr, ok := protobuf.([]interface{})[0].(map[string]interface{})["schema_registry_connection"]; ok {
-			message := protobuf.([]interface{})[0].(map[string]interface{})["message_name"].(string)
+			message := protobuf.([]interface{})[0].(map[string]interface{})["message"].(string)
```
arusahni, line 76: Is this a bugfix? It's not immediately obvious where this is flowing from.

pkg/resources/resource_source_kafka.go
```diff
@@ -32,62 +32,72 @@ var sourceKafkaSchema = map[string]*schema.Schema{
 		ForceNew:    true,
 	},
 	"include_key": {
-		Description: "Include a column containing the Kafka message key.",
+		Description: "Include a column containing the Kafka message key. Deprecated: Use the new materialize_source_table_kafka resource instead.",
+		Deprecated:  "Use the new materialize_source_table_kafka resource instead.",
```
arusahni, line 36:
```suggestion
		Description: "Include a column containing the Kafka message key. Deprecated: Use the new `materialize_source_table_kafka` resource instead.",
		Deprecated:  "Use the new `materialize_source_table_kafka` resource instead.",
```

pkg/resources/resource_source_kafka.go
```diff
@@ -32,62 +32,72 @@ var sourceKafkaSchema = map[string]*schema.Schema{
...
 		Type:        schema.TypeBool,
 		Optional:    true,
 		ForceNew:    true,
 	},
 	"include_key_alias": {
-		Description: "Provide an alias for the key column.",
+		Description: "Provide an alias for the key column. Deprecated: Use the new materialize_source_table_kafka resource instead.",
+		Deprecated:  "Use the new materialize_source_table_kafka resource instead.",
```
arusahni, line 43:
```suggestion
		Description: "Provide an alias for the key column. Deprecated: Use the new `materialize_source_table_kafka` resource instead.",
		Deprecated:  "Use the new `materialize_source_table_kafka` resource instead.",
```

pkg/resources/resource_source_kafka.go
```diff
@@ -32,62 +32,72 @@ var sourceKafkaSchema = map[string]*schema.Schema{
...
 		Type:        schema.TypeString,
 		Optional:    true,
 		ForceNew:    true,
 	},
 	"include_headers": {
-		Description: "Include message headers.",
+		Description: "Include message headers. Deprecated: Use the new materialize_source_table_kafka resource instead.",
+		Deprecated:  "Use the new materialize_source_table_kafka resource instead.",
```
arusahni, line 50:
```suggestion
		Description: "Include message headers. Deprecated: Use the new `materialize_source_table_kafka` resource instead.",
		Deprecated:  "Use the new `materialize_source_table_kafka` resource instead.",
```

pkg/resources/resource_source_kafka.go
```diff
@@ -32,62 +32,72 @@ var sourceKafkaSchema = map[string]*schema.Schema{
...
 		Type:        schema.TypeBool,
 		Optional:    true,
 		ForceNew:    true,
 		Default:     false,
 	},
 	"include_headers_alias": {
-		Description: "Provide an alias for the headers column.",
+		Description: "Provide an alias for the headers column. Deprecated: Use the new materialize_source_table_kafka resource instead.",
+		Deprecated:  "Use the new materialize_source_table_kafka resource instead.",
```
arusahni, line 58:
```suggestion
		Description: "Provide an alias for the headers column. Deprecated: Use the new `materialize_source_table_kafka` resource instead.",
		Deprecated:  "Use the new `materialize_source_table_kafka` resource instead.",
```

pkg/resources/resource_source_kafka.go
```diff
@@ -32,62 +32,72 @@ var sourceKafkaSchema = map[string]*schema.Schema{
...
 		Type:        schema.TypeString,
 		Optional:    true,
 		ForceNew:    true,
 	},
 	"include_partition": {
-		Description: "Include a partition column containing the Kafka message partition",
+		Description: "Include a partition column containing the Kafka message partition. Deprecated: Use the new materialize_source_table_kafka resource instead.",
+		Deprecated:  "Use the new materialize_source_table_kafka resource instead.",
```
arusahni, line 65:
```suggestion
		Description: "Include a partition column containing the Kafka message partition. Deprecated: Use the new `materialize_source_table_kafka` resource instead.",
		Deprecated:  "Use the new `materialize_source_table_kafka` resource instead.",
```

pkg/resources/resource_source_kafka.go
```diff
@@ -32,62 +32,72 @@ var sourceKafkaSchema = map[string]*schema.Schema{
...
 		Type:        schema.TypeBool,
 		Optional:    true,
 		ForceNew:    true,
 	},
 	"include_partition_alias": {
-		Description: "Provide an alias for the partition column.",
+		Description: "Provide an alias for the partition column. Deprecated: Use the new materialize_source_table_kafka resource instead.",
+		Deprecated:  "Use the new materialize_source_table_kafka resource instead.",
```
arusahni, line 72:
```suggestion
		Description: "Provide an alias for the partition column. Deprecated: Use the new `materialize_source_table_kafka` resource instead.",
		Deprecated:  "Use the new `materialize_source_table_kafka` resource instead.",
```

pkg/resources/resource_source_kafka.go
```diff
@@ -32,62 +32,72 @@ var sourceKafkaSchema = map[string]*schema.Schema{
...
 		Type:        schema.TypeString,
 		Optional:    true,
 		ForceNew:    true,
 	},
 	"include_offset": {
-		Description: "Include an offset column containing the Kafka message offset.",
+		Description: "Include an offset column containing the Kafka message offset. Deprecated: Use the new materialize_source_table_kafka resource instead.",
+		Deprecated:  "Use the new materialize_source_table_kafka resource instead.",
```
arusahni, line 79:
```suggestion
		Description: "Include an offset column containing the Kafka message offset. Deprecated: Use the new `materialize_source_table_kafka` resource instead.",
		Deprecated:  "Use the new `materialize_source_table_kafka` resource instead.",
```

pkg/resources/resource_source_kafka.go
```diff
@@ -32,62 +32,72 @@ var sourceKafkaSchema = map[string]*schema.Schema{
...
 		Type:        schema.TypeBool,
 		Optional:    true,
 		ForceNew:    true,
 	},
 	"include_offset_alias": {
-		Description: "Provide an alias for the offset column.",
+		Description: "Provide an alias for the offset column. Deprecated: Use the new materialize_source_table_kafka resource instead.",
+		Deprecated:  "Use the new materialize_source_table_kafka resource instead.",
```
arusahni, line 86:
```suggestion
		Description: "Provide an alias for the offset column. Deprecated: Use the new `materialize_source_table_kafka` resource instead.",
		Deprecated:  "Use the new `materialize_source_table_kafka` resource instead.",
```

pkg/resources/resource_source_kafka.go
```diff
@@ -32,62 +32,72 @@ var sourceKafkaSchema = map[string]*schema.Schema{
...
 		Type:        schema.TypeString,
 		Optional:    true,
 		ForceNew:    true,
 	},
 	"include_timestamp": {
-		Description: "Include a timestamp column containing the Kafka message timestamp.",
+		Description: "Include a timestamp column containing the Kafka message timestamp. Deprecated: Use the new materialize_source_table_kafka resource instead.",
+		Deprecated:  "Use the new materialize_source_table_kafka resource instead.",
```
arusahni, line 93:
```suggestion
		Description: "Include a timestamp column containing the Kafka message timestamp. Deprecated: Use the new `materialize_source_table_kafka` resource instead.",
		Deprecated:  "Use the new `materialize_source_table_kafka` resource instead.",
```

pkg/resources/resource_source_kafka.go
```diff
@@ -32,62 +32,72 @@ var sourceKafkaSchema = map[string]*schema.Schema{
...
 		Type:        schema.TypeBool,
 		Optional:    true,
 		ForceNew:    true,
 	},
 	"include_timestamp_alias": {
-		Description: "Provide an alias for the timestamp column.",
+		Description: "Provide an alias for the timestamp column. Deprecated: Use the new materialize_source_table_kafka resource instead.",
+		Deprecated:  "Use the new materialize_source_table_kafka resource instead.",
```
arusahni, line 100:
```suggestion
		Description: "Provide an alias for the timestamp column. Deprecated: Use the new `materialize_source_table_kafka` resource instead.",
		Deprecated:  "Use the new `materialize_source_table_kafka` resource instead.",
```

pkg/resources/resource_source_kafka.go
```diff
@@ -96,7 +106,8 @@ var sourceKafkaSchema = map[string]*schema.Schema{
 	"key_format":   FormatSpecSchema("key_format", "Set the key format explicitly.", false),
 	"value_format": FormatSpecSchema("value_format", "Set the value format explicitly.", false),
 	"envelope": {
-		Description: "How Materialize should interpret records (e.g. append-only, upsert)..",
+		Description: "How Materialize should interpret records (e.g. append-only, upsert). Deprecated: Use the new materialize_source_table_kafka resource instead.",
+		Deprecated:  "Use the new materialize_source_table_kafka resource instead.",
```
arusahni, line 110:
```suggestion
		Description: "How Materialize should interpret records (e.g. append-only, upsert). Deprecated: Use the new `materialize_source_table_kafka` resource instead.",
		Deprecated:  "Use the new `materialize_source_table_kafka` resource instead.",
```

pkg/resources/resource_source_mysql.go
```diff
@@ -27,19 +27,22 @@ var sourceMySQLSchema = map[string]*schema.Schema{
 		ForceNew:    true,
 	}),
 	"ignore_columns": {
-		Description: "Ignore specific columns when reading data from MySQL. Can only be updated in place when also updating a corresponding `table` attribute.",
+		Description: "Ignore specific columns when reading data from MySQL. Can only be updated in place when also updating a corresponding `table` attribute. Deprecated: Use the new materialize_source_table_mysql resource instead.",
+		Deprecated:  "Use the new materialize_source_table_mysql resource instead.",
```
arusahni, line 31:
```suggestion
		Description: "Ignore specific columns when reading data from MySQL. Can only be updated in place when also updating a corresponding `table` attribute. Deprecated: Use the new `materialize_source_table_mysql` resource instead.",
		Deprecated:  "Use the new `materialize_source_table_mysql` resource instead.",
```

pkg/resources/resource_source_mysql.go
```diff
@@ -27,19 +27,22 @@ var sourceMySQLSchema = map[string]*schema.Schema{
...
 		Type:        schema.TypeList,
 		Elem:        &schema.Schema{Type: schema.TypeString},
 		Optional:    true,
 	},
 	"text_columns": {
-		Description: "Decode data as text for specific columns that contain MySQL types that are unsupported in Materialize. Can only be updated in place when also updating a corresponding `table` attribute.",
+		Description: "Decode data as text for specific columns that contain MySQL types that are unsupported in Materialize. Can only be updated in place when also updating a corresponding `table` attribute. Deprecated: Use the new materialize_source_table_mysql resource instead.",
+		Deprecated:  "Use the new materialize_source_table_mysql resource instead.",
```
arusahni, line 38:
```suggestion
		Description: "Decode data as text for specific columns that contain MySQL types that are unsupported in Materialize. Can only be updated in place when also updating a corresponding `table` attribute. Deprecated: Use the new `materialize_source_table_mysql` resource instead.",
		Deprecated:  "Use the new `materialize_source_table_mysql` resource instead.",
```

pkg/resources/resource_source_mysql.go
```diff
@@ -27,19 +27,22 @@ var sourceMySQLSchema = map[string]*schema.Schema{
...
 		Type:        schema.TypeList,
 		Elem:        &schema.Schema{Type: schema.TypeString},
 		Optional:    true,
 	},
 	"table": {
-		Description: "Specify the tables to be included in the source. If not specified, all tables are included.",
+		Description: "Specify the tables to be included in the source. Deprecated: Use the new materialize_source_table_mysql resource instead.",
+		Deprecated:  "Use the new materialize_source_table_mysql resource instead.",
```
arusahni, line 45:
```suggestion
		Description: "Specify the tables to be included in the source. Deprecated: Use the new `materialize_source_table_mysql` resource instead.",
		Deprecated:  "Use the new `materialize_source_table_mysql` resource instead.",
```

pkg/resources/resource_source_mysql.go
```diff
@@ -76,6 +79,13 @@ var sourceMySQLSchema = map[string]*schema.Schema{
 			},
 		},
 	},
+	"all_tables": {
+		Description: "Include all tables in the source. If `table` is specified, this will be ignored.",
+		Deprecated:  "Use the new materialize_source_table_mysql resource instead.",
```
arusahni, line 84:
```suggestion
		Deprecated:  "Use the new `materialize_source_table_mysql` resource instead.",
```

pkg/resources/resource_source_postgres.go
```diff
@@ -33,14 +33,17 @@ var sourcePostgresSchema = map[string]*schema.Schema{
 		ForceNew:    true,
 	},
 	"text_columns": {
-		Description: "Decode data as text for specific columns that contain PostgreSQL types that are unsupported in Materialize. Can only be updated in place when also updating a corresponding `table` attribute.",
+		Description: "Decode data as text for specific columns that contain PostgreSQL types that are unsupported in Materialize. Can only be updated in place when also updating a corresponding `table` attribute. Deprecated: Use the new materialize_source_table_postgres resource instead.",
+		Deprecated:  "Use the new materialize_source_table_postgres resource instead.",
```
arusahni, line 37:
```suggestion
		Description: "Decode data as text for specific columns that contain PostgreSQL types that are unsupported in Materialize. Can only be updated in place when also updating a corresponding `table` attribute. Deprecated: Use the new `materialize_source_table_postgres` resource instead.",
		Deprecated:  "Use the new `materialize_source_table_postgres` resource instead.",
```

pkg/resources/resource_source_postgres.go
```diff
@@ -33,14 +33,17 @@ var sourcePostgresSchema = map[string]*schema.Schema{
...
 		Type:        schema.TypeList,
 		Elem:        &schema.Schema{Type: schema.TypeString},
 		Optional:    true,
 	},
 	"table": {
-		Description: "Creates subsources for specific tables in the Postgres connection.",
+		Description: "Creates subsources for specific tables in the Postgres connection. Deprecated: Use the new materialize_source_table_postgres resource instead.",
+		Deprecated:  "Use the new materialize_source_table_postgres resource instead.",
```
arusahni, line 44:
```suggestion
		Description: "Creates subsources for specific tables in the Postgres connection. Deprecated: Use the new `materialize_source_table_postgres` resource instead.",
		Deprecated:  "Use the new `materialize_source_table_postgres` resource instead.",
```

pkg/resources/resource_source_table_mysql.go
```diff
@@ -0,0 +1,243 @@
...
+	},
+	"text_columns": {
+		Description: "Columns to be decoded as text.",
+		Type:        schema.TypeList,
+		Elem:        &schema.Schema{Type: schema.TypeString},
+		Optional:    true,
+		ForceNew:    true,
+	},
+	"exclude_columns": {
+		Description: "Exclude specific columns when reading data from MySQL. The option used to be called `ignore_columns`.",
```
arusahni, line 46:
```suggestion
		Description: "Exclude specific columns when reading data from MySQL. This option used to be called `ignore_columns`.",
```

arusahni (approved):
Don't let the ~51 comments scare you off -- they're mostly nits. Great job on this!!!

pkg/materialize/source_table_kafka.go
```diff
@@ -0,0 +1,389 @@
...
+		q := strings.Builder{}
+		var options []string
+
+		// Format
+		if b.format.Avro != nil {
+			if b.format.Avro.SchemaRegistryConnection.Name != "" {
+				options = append(options, fmt.Sprintf(`FORMAT AVRO USING CONFLUENT SCHEMA REGISTRY CONNECTION %s`, QualifiedName(b.format.Avro.SchemaRegistryConnection.DatabaseName, b.format.Avro.SchemaRegistryConnection.SchemaName, b.format.Avro.SchemaRegistryConnection.Name)))
+			}
+			if b.format.Avro.KeyStrategy != "" {
+				options = append(options, fmt.Sprintf(`KEY STRATEGY %s`, b.format.Avro.KeyStrategy))
```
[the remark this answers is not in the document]
bobbyiliev: Here we would not need to escape this is a keyword. If we were to escape it it would result in the following error on the Materialize side:

```
Expected one of ID or LATEST or INLINE, found string literal "INLINE"
```

We do however have validation in place on the terraform schema side already, so a user will only be able to specify one of the following values: https://github.com/MaterializeInc/terraform-provider-materialize/blob/68319193d1576420971d5e4cb95b60e8ab13fab3/pkg/resources/config.go#L58-L62

pkg/materialize/source_table_kafka.go
```diff
@@ -0,0 +1,389 @@
...
+			}
+			if b.format.Avro.ValueStrategy != "" {
+				options = append(options, fmt.Sprintf(`VALUE STRATEGY %s`, b.format.Avro.ValueStrategy))
```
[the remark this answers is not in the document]
bobbyiliev: Same as above:

Here we would not need to escape this is a keyword. If we were to escape it it would result in the following error on the Materialize side:

```
Expected one of ID or LATEST or INLINE, found string literal "INLINE"
```

We do however have validation in place on the terraform schema side already, so a user will only be able to specify one of the following values: https://github.com/MaterializeInc/terraform-provider-materialize/blob/68319193d1576420971d5e4cb95b60e8ab13fab3/pkg/resources/config.go#L58-L62

pkg/materialize/format_specs.go
bobbyiliev: Yep, seems like it has been there for a while 🤦 
https://github.com/MaterializeInc/terraform-provider-materialize/blob/68319193d1576420971d5e4cb95b60e8ab13fab3/pkg/resources/schema.go#L212-L217

pkg/resources/resource_source_table_postgres.go
```diff
@@ -0,0 +1,228 @@
...
+			b.Drop()
+			return diag.FromErr(err)
+		}
+	}
+
+	// Handle comments
+	if v, ok := d.GetOk("comment"); ok {
+		comment := materialize.NewCommentBuilder(metaDb, o)
+		if err := comment.Object(v.(string)); err != nil {
+			log.Printf("[DEBUG] resource failed comment, dropping object: %s", o.Name)
```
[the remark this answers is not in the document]
bobbyiliev: Yep, those should be ok, they will only show if the `TF_LOG` env var is set to `DEBUG` and can be helpful for debugging, [similar to how other providers do this](https://github.com/search?q=repo%3ASnowflake-Labs%2Fterraform-provider-snowflake+%5BDEBUG%5D&type=code).

But this is actually a very good point, going through the TF docs, it looks like that they now suggest using their tflog package instead of the good old print + `[log_level]`. I've created a tracking issue for this so we could look into it further and possibly plan the work for this! https://github.com/MaterializeInc/terraform-provider-materialize/issues/664

docs/data-sources/source_reference.md
```diff
@@ -0,0 +1,53 @@
+---
+# generated by https://github.com/hashicorp/terraform-plugin-docs
+page_title: "materialize_source_reference Data Source - terraform-provider-materialize"
+subcategory: ""
+description: |-
```
rjobanp, line 5: Can we populate this to explain that these are 'available' source references? Such that these expose all the possible upstream references that this source can create a table for, not necessarily all the references it is already ingesting

docs/guides/materialize_source_table.md
```diff
@@ -0,0 +1,245 @@
...
+  Guide for migrating to the new materialize_source_table_{source_type} resources.
+---
+
+# Source versioning: migrating to `materialize_source_table_{source_type}` Resource
+
+In previous versions of the Materialize Terraform provider, source tables were defined within the source resource itself and were considered subsources of the source rather than separate entities.
+
+This guide will walk you through the process of migrating your existing source table definitions to the new `materialize_source_table_{source_type}` resource.
+
+For each source type (e.g., MySQL, Postgres, etc.), you will need to create a new `materialize_source_table_{source_type}` resource for each table that was previously defined within the source resource. This ensures that the tables are preserved during the migration process. For Kafka sources, you will need to create at least one `materialize_source_table_kafka` table to hold data for the kafka topic.
```
rjobanp, line 16:
```suggestion
For each MySQL and Postgres source, you will need to create a new `materialize_source_table_{source_type}` resource for each table that was previously defined within the source resource. This ensures that the tables are preserved during the migration process. For Kafka sources, you will need to create a `materialize_source_table_kafka` table with the same name as the kafka source to contain the data for the kafka topic.
```

docs/guides/materialize_source_table.md
```diff
@@ -0,0 +1,245 @@
...
+subcategory: ""
+description: |-
+  Guide for migrating to the new materialize_source_table_{source_type} resources.
+---
+
+# Source versioning: migrating to `materialize_source_table_{source_type}` Resource
+
+In previous versions of the Materialize Terraform provider, source tables were defined within the source resource itself and were considered subsources of the source rather than separate entities.
+
+This guide will walk you through the process of migrating your existing source table definitions to the new `materialize_source_table_{source_type}` resource.
```
rjobanp, line 14: I'm confused on whether this guide is meant to explain how to migrate using just the manual process (which is outlined below and looks correct), or also to explain how to reconcile your terraform configuration with the results of the 'automatic' migration process that we would do for a customer with a catalog migration. Specifically, the paragraph below sounds like it is explaining what to do to handle the automatic migration process, but it doesn't explain that the manual process would then be unnecessary.

It might be simpler if this guide just explained the manual process and we had a separate one to explain the 'automatic' process. Or we could just assume any terraform user would only do the manual process. @morsapaes thoughts?

docs/guides/materialize_source_table.md
```diff
@@ -0,0 +1,245 @@
...
+}
+```
+
+This `ignore_changes` block tells Terraform to ignore changes to these attributes during subsequent applies, preventing Terraform from trying to update these values based on incomplete information from the state.
+
+After importing, you may need to manually update these ignored attributes in your Terraform configuration to match the actual state in Materialize.
+
+## Future Improvements
+
+The Kafka and Webhooks sources are currently being implemented. Once these changes, the migration process will be updated to include them.
```
rjobanp, line 245:
```suggestion
Webhooks sources have not yet been migrated to the new model. Once this changes, the migration process will be updated to include them.
```

docs/resources/source_table_kafka.md
```diff
@@ -0,0 +1,383 @@
...
+```
+
+<!-- schema generated by tfplugindocs -->
+## Schema
+
+### Required
+
+- `name` (String) The identifier for the source table.
+- `source` (Block List, Min: 1, Max: 1) The source this table is created from. (see [below for nested schema](#nestedblock--source))
+- `topic` (String) The name of the Kafka topic in the Kafka cluster.
```
rjobanp, line 71: this is actually no longer necessary to include for Kafka sources -- if the `(REFERENCE ..)` option in the statement is omitted it will still work since there is only one possible kafka topic that the table can reference

pkg/materialize/source_reference.go
```diff
@@ -0,0 +1,76 @@
...
+	}
+
+	return s, nil
+}
+
+func ListSourceReferences(conn *sqlx.DB, sourceId string) ([]SourceReferenceParams, error) {
+	p := map[string]string{
+		"sr.source_id": sourceId,
+	}
+	q := sourceReferenceQuery.QueryPredicate(p)
```
rjobanp, line 68: this would be a bonus, but there is a PR to implement a `REFRESH SOURCE REFERENCES <source>` statement here https://github.com/MaterializeInc/materialize/pull/29923 and we could automatically run that before listing the references which would mean we always get the most 'up to date' view of them

pkg/materialize/source_table.go
```diff
@@ -0,0 +1,192 @@
...
+	q.WriteString(fmt.Sprintf(`CREATE TABLE %s`, b.QualifiedName()))
+	q.WriteString(fmt.Sprintf(` FROM SOURCE %s`, b.source.QualifiedName()))
+	q.WriteString(` (REFERENCE `)
+
+	if b.upstreamSchemaName != "" {
+		q.WriteString(fmt.Sprintf(`%s.`, QuoteIdentifier(b.upstreamSchemaName)))
+	}
+	q.WriteString(QuoteIdentifier(b.upstreamName))
+
+	q.WriteString(")")
```
rjobanp, line 166: mentioned above that this can be made optional for kafka and single-output load generator sources

pkg/materialize/source_table_kafka.go
```diff
@@ -0,0 +1,389 @@
...
+var sourceTableKafkaQuery = `
+	SELECT
+		mz_tables.id,
+		mz_tables.name,
+		mz_schemas.name AS schema_name,
+		mz_databases.name AS database_name,
+		mz_sources.name AS source_name,
+		source_schemas.name AS source_schema_name,
+		source_databases.name AS source_database_name,
+		mz_kafka_source_tables.topic AS upstream_table_name,
```
rjobanp, line 23: I have a PR open now to add the `envelope`, `key_format` and `value_format` columns to this table https://github.com/MaterializeInc/materialize/pull/30076

rjobanp (approved):
looks solid! Just a few minor comments on some things that have been updated, and one about the migration guide

pkg/materialize/source_table_kafka.go
bobbyiliev: Neat! I created a tracking issue for this for the moment: https://github.com/MaterializeInc/terraform-provider-materialize/issues/665

Should be ok to handle in a follow up PR later on!

pkg/materialize/source_reference.go
bobbyiliev: Just implemented this! 
https://github.com/MaterializeInc/terraform-provider-materialize/blob/147606a7ddd52a7f50a4eee6a9d1cb31030afd7f/pkg/materialize/source_reference.go#L65-L69

docs/guides/materialize_source_table.md
bobbyiliev: Good point! I've removed some of the confusing text so that this should focus on the manual migration!

Happy to work on a follow up migration guide for the automated migration if we decide to do that!

pkg/provider/acceptance_source_table_kafka_test.go
```diff
@@ -0,0 +1,230 @@
...
+		Steps: []resource.TestStep{
+			{
+				Config: testAccSourceTableKafkaBasicResource(nameSpace),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckSourceTableExists("materialize_source_table_kafka.test_kafka"),
+					resource.TestMatchResourceAttr("materialize_source_table_kafka.test_kafka", "id", terraformObjectIdRegex),
+					resource.TestCheckResourceAttr("materialize_source_table_kafka.test_kafka", "name", nameSpace+"_table_kafka"),
+					resource.TestCheckResourceAttr("materialize_source_table_kafka.test_kafka", "database_name", "materialize"),
+					resource.TestCheckResourceAttr("materialize_source_table_kafka.test_kafka", "schema_name", "public"),
+					// resource.TestCheckResourceAttr("materialize_source_table_kafka.test_kafka", "qualified_sql_name", fmt.Sprintf(`"materialize"."public"."%s_table_kafka"`, nameSpace)),
```
ParkMyCar, line 28: left behind?

ParkMyCar (approved):
Really just skimmed the entire PR but I think it all looks good! Biggest feedback would be following up on quoting identifiers that Aru commented on, but if there aren't any concerns there I would say we're good to go!

PR #653, opened 2024-09-09
Pin Frontegg mock to stable version

bobbyiliev:
Switching the Frontegg mock and certs version to v0.116.0 which includes the latest Frontegg mock changes and refactor.


arusahni (approved):
Nice!

PR #657, opened 2024-09-27
Set transaction isolation to strict serializable

### LINKED ISSUE #656: Set transaction isolation to strict for new connections
### Feature request

Recently @matthelm observed an unexpected behaviour when creating new resources. 

The problem was caused when the transaction isolation in Materialize was set to `serializable` causing the read function to intermittently fail right after a create statement is executed.

The suggested fix is to explicitly run `SET transaction_isolation = 'strict serializable'` whenever we create a new connection in Terraform.

bobbyiliev:
Explicitly set the transaction isolation level to 'strict serializable' for each db client.

Fixes #656

### DIFF
--- a/compose.yaml
+++ b/compose.yaml
@@ -48,6 +48,7 @@ services:
       - --system-parameter-default=max_clusters=100
       - --system-parameter-default=max_sources=100
       - --system-parameter-default=max_aws_privatelink_connections=10
+      - --system-parameter-default=transaction_isolation=serializable
       - --all-features
     environment:
       MZ_NO_TELEMETRY: 1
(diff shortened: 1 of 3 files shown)


compose.yaml
```diff
@@ -48,6 +48,7 @@ services:
       - --system-parameter-default=max_clusters=100
       - --system-parameter-default=max_sources=100
       - --system-parameter-default=max_aws_privatelink_connections=10
+      - --system-parameter-default=transaction_isolation=serializable
```
bobbyiliev, line 51: Setting the transaction isolation level to serializable for the second region, that way we will make sure that even if the default value is not set to strict serializable, everything will still work as the provider will explicitly set the level to strict serializable.

matthelm:
Thanks for your help!

arusahni (approved):
Nice!

compose.yaml
benesch: Smart!

benesch (approved):

PR #660, opened 2024-10-02
Set transaction_isolation as conneciton option

bobbyiliev:
Decided to refactor this a bit and set the transaction isolation to strict serializable via the PG options rather than actually running the SET command. This makes things easier for the unit tests.

### DIFF
--- a/pkg/clients/db_client.go
+++ b/pkg/clients/db_client.go
@@ -33,6 +33,8 @@ func NewDBClient(host, user, password string, port int, database, application_na
 }
 
 func buildConnectionString(host, user, password string, port int, database, sslmode, application_name string) string {
+	options := "--transaction_isolation=strict\\ serializable"
+
 	url := &url.URL{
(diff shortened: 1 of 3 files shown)


pkg/clients/db_client.go
```diff
@@ -33,6 +33,8 @@ func NewDBClient(host, user, password string, port int, database, application_na
 }
 
 func buildConnectionString(host, user, password string, port int, database, sslmode, application_name string) string {
+	options := "--transaction_isolation=strict\\ serializable"
```
bobbyiliev, line 36: Embarrassed to admit how long it took me to get this syntax correct 😁

arusahni: The backslashes will continue until morale improves :laughing:

arusahni (approved):
I see a little value in having the explicit `SET` call, but if this makes things easier for testing, that's great!

PR #671, opened 2024-11-17
Add bootstrap-builtin-analytics-cluster-replica-size

bobbyiliev:
Adding `--bootstrap-builtin-analytics-cluster-replica-size=3xsmall` in order to fix the following panic with the latest materialzied image:

```
 Connection could not be recycled: Connection closed    
thread 'main' panicked at src/environmentd/src/bin/environmentd/main.rs:635:9:
environmentd: fatal: unknown cluster replica size '1'
   8: rust_begin_unwind
             at ./rustc/3f5fd8dd41153bc5fdca9427e9e05be2c767ba23/library/std/src/panicking.rs:652:5
   9: core::panicking::panic_fmt
             at ./rustc/3f5fd8dd41153bc5fdca9427e9e05be2c767ba23/library/core/src/panicking.rs:72:14
  10: environmentd::main
             at src/environmentd/src/bin/environmentd/main.rs:635:9
  11: <fn() as core::ops::function::FnOnce<()>>::call_once
             at ./rustc/3f5fd8dd41153bc5fdca9427e9e05be2c767ba23/library/core/src/ops/function.rs:250:5
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
```


compose.yaml
```diff
@@ -1,6 +1,4 @@
 ---
-version: "3.9"
```
bobbyiliev, line 2: This is deprecated and no longer needed in compose files

pkg/materialize/cluster.go
```diff
@@ -167,7 +167,8 @@ func (b *ClusterBuilder) GenerateClusterOptions() string {
 		p = append(p, i)
 	}
 
-	if b.disk {
+	// Only add DISK to the quiery builder if it's enabled AND size doesn't end in either "cc" or "C"
+	if b.disk && !strings.HasSuffix(b.size, "cc") && !strings.HasSuffix(b.size, "C") {
```
bobbyiliev, line 171: This was working fine in the past as the disk option was silently ignored in Materialize for cc and C clusters. However this no longer seems to be the case so we need to explicitly skip it in the query builder.

benesch (approved):

def- (approved):

PR #674, opened 2024-11-27
Support self-hosted instances

bobbyiliev:
Now users can configure the provider like this:

```hcl
# Self-hosted configuration
provider "materialize" {
  host     = "localhost"      # Required for self-hosted mode
  port     = 6875             # Optional, defaults to 6875
  database = "materialize"    # Optional, defaults to materialize
  username = "materialize"    # Optional, defaults to materialize
  password = ""               # Optional
  sslmode  = "disable"        # Optional, defaults to require
}

# SaaS configuration (unchanged)
provider "materialize" {
  password       = "materialize_password"
  default_region = "aws/us-east-1"
}
```


pkg/provider/provider.go
```diff
@@ -63,6 +63,24 @@ func Provider(version string) *schema.Provider {
...
+				Type:        schema.TypeString,
+				Optional:    true,
+				Description: "The Materialize host. Can also come from the `MZ_HOST` environment variable.",
+				DefaultFunc: schema.EnvDefaultFunc("MZ_HOST", nil),
+			},
+			"port": {
+				Type:        schema.TypeInt,
+				Optional:    true,
+				Description: "The Materialize port. Can also come from the `MZ_PORT` environment variable.",
+				DefaultFunc: schema.EnvDefaultFunc("MZ_PORT", 6875),
```
alex-hunt-materialize, line 76: For self-hosted, do we want to point at the external SQL address, or the internal one? The external one doesn't allow login as `mz_system`, and other users have limited permissions.

Also, nitpick, it might be good to name this in a way that is clearly a SQL port, as we also have HTTP(S) ports.

alex-hunt-materialize (commented):
Code mostly looks good to me. Just one question and nitpick.

integration/self_hosted/cluster.tf
```diff
@@ -0,0 +1,66 @@
+resource "materialize_cluster" "cluster" {
```
bobbyiliev, line 1: The large number of line changes in this PR is due to all of the files in the `integration/self_hosted` directory. Those are just standard tests. If reviewing this PR, focus on the changes in the `pkg` directory.

pkg/provider/provider.go
```diff
@@ -156,6 +174,56 @@ func Provider(version string) *schema.Provider {
...
 
 func providerConfigure(ctx context.Context, d *schema.ResourceData, version string) (interface{}, diag.Diagnostics) {
+	// Check for self-hosted configuration
+	if host := d.Get("host").(string); host != "" {
+		log.Printf("[DEBUG] Configuring self-hosted provider")
+		return configureSelfHosted(ctx, d, version)
+	}
+	log.Printf("[DEBUG] Configuring SaaS provider")
+	return configureSaaS(ctx, d, version)
+}
```
bobbyiliev, line 184: This is where we determine if we should use the self-hosted authentication flow or the SaaS authentication flow.

integration/self_hosted/main.tf
```diff
@@ -0,0 +1,16 @@
...
+    materialize = {
+      source = "materialize.com/devex/materialize"
+    }
+  }
+}
+
+provider "materialize" {
+  host     = "materialized"
+  port     = 6877
+  database = "materialize"
```
alex-hunt-materialize, line 12: Missed this one.

bobbyiliev: This is only for the internal integration tests which are using the Docker emulator. It allows me to test some resources that the `materialize` user does not have the rights to, like setting system parameters for example:

```hcl
resource "materialize_system_parameter" "default_cluster" {
  name  = "max_tables"
  value = "2000"
}
```

This will not be exposed to the end users and the orchestratord.

alex-hunt-materialize (approved):

PR #690, opened 2025-01-14
Add support for service accounts to provider auth

bobbyiliev:
Fixes the following when using service accounts for the provider auth:

```sh
Unable to create Frontegg client: failed to get token: email claim not found in token
```

Depends on: https://github.com/MaterializeInc/cloud/pull/10632

### DIFF
--- a/pkg/clients/frontegg_client.go
+++ b/pkg/clients/frontegg_client.go
@@ -127,7 +127,20 @@ func getToken(ctx context.Context, password string, endpoint string) (string, st
 
 	email, ok := claims["email"].(string)
 	if !ok {
-		return "", "", time.Time{}, errors.New("email claim not found in token")
+		// If email is not present (service account case), use metadata.user or sub as identifier
+		if metadata, hasMetadata := claims["metadata"].(map[string]interface{}); hasMetadata {
+			if user, hasUser := metadata["user"].(string); hasUser {
+				email = user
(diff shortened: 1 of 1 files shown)


pkg/clients/frontegg_client.go
```diff
@@ -127,7 +127,20 @@ func getToken(ctx context.Context, password string, endpoint string) (string, st
...
 	if !ok {
-		return "", "", time.Time{}, errors.New("email claim not found in token")
+		// If email is not present (service account case), use metadata.user or sub as identifier
+		if metadata, hasMetadata := claims["metadata"].(map[string]interface{}); hasMetadata {
+			if user, hasUser := metadata["user"].(string); hasUser {
+				email = user
+			}
+		}
+
+		if email == "" {
```
SangJunBak, line 137: Is `email` guaranteed to be an empty string when `claims["email"]` doesn't return `ok` as `true`? If not, wondering if we should add an `email = ""` right before `if metadata, hasMetadata := claims["metadata"].(map[string]interface{}); hasMetadata {`

SangJunBak (approved):

pkg/clients/frontegg_client.go
bobbyiliev: Ah yes, this is a good point! Thanks Jun!

Issue #686, opened 2024-12-20
Add support for `CREATE TABLE ... FROM WEBHOOK`


bobbyiliev:
### Feature request

https://github.com/MaterializeInc/materialize/pull/30821

Issue #602, opened 2024-07-11
Implement SQL parsing for `CREATE TABLE .. FROM SOURCE`


bobbyiliev:
https://github.com/MaterializeInc/materialize/pull/28125

PR #692, opened 2025-01-17
[WIP] Source versioning

bobbyiliev:
Currently on-hold but this is the initial implementation for the source versioning refactor as per #646 

The main changes to consider:
- Marking the `table` attribute as optional and deprecated for both the MySQL and Postgres sources
- Introduced a new `all_tables` bool attribute for the MySQL and the Loadgen sources, as in the past this was defaulting always using `FOR ALL TABLES` in the load gen sources (auction, marketing, tpch) and in the MySQL case whenever no `table` blocks were defined, we defaulted to `FOR ALL TABLES`. This `all_tables` bool attribute allows us to create sources without any tables defined as per the source versioning work
- Introducing the new `materialize_source_table_{mysql|postgres|load_generator}` resource which allows us to do `CREATE TABLE ... FROM SOURCE ...`

Things that are still pending: https://github.com/MaterializeInc/terraform-provider-materialize/issues/646

Currently missing tests for the webhook source.

Fixes https://github.com/MaterializeInc/terraform-provider-materialize/issues/686
Fixes https://github.com/MaterializeInc/terraform-provider-materialize/issues/602

Covers most of the tasks under https://github.com/MaterializeInc/terraform-provider-materialize/issues/646


bobbyiliev:
Closing this PR in favour of: https://github.com/MaterializeInc/terraform-provider-materialize/pull/693

PR #727, opened 2025-07-25
Add external id attr to aws conn

### LINKED ISSUE #726: `materialize_connection_aws` external_id attribute
### Feature request

If `assume_role_arn` is passed to a `materialize_connection_aws` resource, the next step would intuitively be to use the external ID to set up an assume role policy. As far as I can tell, we can't do this; we'd need to query `mz_internal.mz_aws_connections` for the external ID value, and I don't believe there's a way to do that in Terraform.
Could a read-only `external_id` attribute be added to the `materialize_connection_aws` resource?

My specific use case is translating step 2 of these docs into a Terraform module: https://materialize.com/docs/serve-results/sink/s3/

bobbyiliev:
Fixes #726 

For the PR review check out the 7899ed4 commit. The second commit is just the auto generated docs due to a new version in the TF docs package.

### DIFF
--- a/docs/resources/app_password.md
+++ b/docs/resources/app_password.md
@@ -59,6 +59,8 @@ resource "materialize_app_password" "example_app_password" {
 
 Import is supported using the following syntax:
 
+The [`terraform import` command](https://developer.hashicorp.com/terraform/cli/commands/import) can be used, for example:
+
 ```shell
 # App passwords can be imported using the app password id:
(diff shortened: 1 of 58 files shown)


SangJunBak:
back from the dead

bobbyiliev:
> back from the dead

It’s 1am why you reviewing PRs 😂

Issue #724, opened 2025-07-11
Try to create `materialize_app_password` and got `role not found: Member`


jasonsun9406:
### What version of the Materialize Terraform provider are you using?

v1.12.2

### What version of the Terraform CLI are you using?

v1.12.2

### What version of Materialize are you using?

v0.149.0

## Issues
Got `role not found: Member` when using `materialize_app_password` to create service account app password. Even if I used my account that has Organization Admin permission

## Steps to reproduce
1. I first created a personal app password for myself as I have Organization Admin permission
2. Use the personal app password to setup provider
3. Follow this [example](https://registry.terraform.io/providers/MaterializeInc/materialize/latest/docs/resources/app_password) to deploy an service account app password 

Complete code
```
terraform {
  required_version = ">= 1.4.0, < 2.0.0"

  required_providers {

    materialize = {
      source  = "MaterializeInc/materialize"
      version = "0.8.18"
    }

  }
}
provider "materialize" {
  host     = "{{ host }}" 
  username = "{{ username }}"
  password = "{{ password  }}"
}

resource "materialize_role" "production_dashboard" {
  name = "svc_production_dashboard"
}
resource "materialize_app_password" "production_dashboard_app_password" {
  name  = "production_dashboard_app_password"
  type  = "service"
  user  = materialize_role.production_dashboard.name
  roles = ["Member"]
}
```

Here is the error message I got
<img width="905" height="210" alt="Image" src="https://github.com/user-attachments/assets/61706bb1-40dd-474c-a578-0fde39e1708f" />

PR #729, opened 2025-07-28
fix: Add clear error messages for Frontegg resources in self-hosted mode

bobbyiliev:
## Problem
Users configuring the provider for self-hosted mode were getting confusing errors like "role not found: Member" when trying to use `materialize_app_password` and other Frontegg-dependent resources.

Reported in #724 

## Solution
- Added `ValidateSaaSOnly()` helper that detects self-hosted mode
- Applied validation to all 13 Frontegg-dependent resources + data sources
- Improved provider documentation and examples with clear mode distinctions
- Added better test coverage

## Impact
- Clear, actionable error messages instead of confusing "role not found" errors
- Users immediately understand SaaS vs self-hosted resource limitations
- Better developer experience with upfront documentation about configuration modes

**Before:** `Error: role not found: Member`
**After:** `Error: materialize_app_password is only available in Materialize Cloud (SaaS) environments. You are currently using self-hosted mode...`

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -12,13 +12,29 @@ This repository contains a Terraform provider for managing resources in a [Mater
 Configure the provider by adding the following block to your Terraform project:
 
 ```terraform
-# Configuration-based authentication
+# =============================================================================
+# Materialize Cloud (SaaS) Configuration
+# =============================================================================
+# Use this configuration for Materialize Cloud environments.
+# This provides access to ALL provider resources including:
+# - App passwords, users, SSO, SCIM resources
+# - All database resources (clusters, sources, sinks, etc.)
+#
 provider "materialize" {
   password       = var.materialize_password # optionally use MZ_PASSWORD env var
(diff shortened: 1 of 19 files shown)


docs/index.md
```diff
@@ -12,13 +12,29 @@ This repository contains a Terraform provider for managing resources in a [Mater
...
-# Self-hosted Materialize authentication
+# =============================================================================
+# Self-Hosted Materialize Configuration
+# =============================================================================
+# Use this configuration for self-hosted Materialize instances.
+# 
+# ⚠️  IMPORTANT LIMITATIONS:
+# - Frontegg-dependent resources are NOT available (app passwords, users, SSO, SCIM)
+# - Only database resources are available (clusters, sources, sinks, schemas, etc.)
+# - No organization or identity management features
```
jubrad, line 36: Interesting... do we support `create role ... with password, superuser...` in this terraform? That might be a viable callout here for identity management. If we don't, that might be a reasonable feature to add in the future.

jubrad (approved):

PR #739, opened 2025-09-05
Fix failing disk tests

### LINKED ISSUE #738: `materialize_cluster`: always reports as if there will be updated in-place
### What version of the Materialize Terraform provider are you using?

`latest`

### What version of the Terraform CLI are you using?

v1.9.8

### What version of Materialize are you using?

v0.155.0

### What is the issue?

Tests are now failing with the latest Materialize Docker image because of the following:

```
    acceptance_cluster_test.go:251: Step 1/2 error: After applying this test step, the non-refresh plan was not empty.
        stdout:
        
        
        Terraform used the selected providers to generate the following execution
        plan. Resource actions are indicated with the following symbols:
          ~ update in-place
        
        Terraform will perform the following actions:
        
          # materialize_cluster.test will be updated in-place
          ~ resource "materialize_cluster" "test" {
              ~ disk                    = true -> false
                id                      = "aws/us-east-1:id:u28"
                name                    = "hrlkawnsuu"
                # (9 unchanged attributes hidden)
            }
        
```

bobbyiliev:
Fixes #738


pkg/resources/resource_cluster.go
```diff
@@ -458,7 +445,8 @@ func clusterImport(ctx context.Context, d *schema.ResourceData, meta interface{}
 	d.Set("ownership_role", s.OwnerName.String)
 	d.Set("replication_factor", s.ReplicationFactor.Int64)
 	d.Set("size", s.Size.String)
-	d.Set("disk", s.Disk.Bool)
+	// Disk is always enabled for all clusters now (deprecated feature)
+	d.Set("disk", true)
```
jubrad, line 449: do we even need to set this?

pkg/resources/resource_cluster.go
```diff
@@ -174,7 +173,8 @@ func clusterRead(ctx context.Context, d *schema.ResourceData, meta interface{})
 		return diag.FromErr(err)
 	}
 
-	if err := d.Set("disk", s.Disk.Bool); err != nil {
+	// The disk attr is deprecated and is not configurable
+	if err := d.Set("disk", true); err != nil {
 		return diag.FromErr(err)
 	}
```
jubrad, line 179: Do we need to set this at all?

jubrad (approved):
LG, I had some questions, but it seems like the disk attr will get removed at some point in the future so they can be ignored.

pkg/resources/resource_cluster.go
bobbyiliev: I think we still need to keep this around for backward compatibility, otherwise it might break existing projects. That said, we can clean it up once we decide to do a minor version bump.

PR #757, opened 2025-11-07
Add write-only value_wo on materialize_secret

djahandarie:
# Background

Terraform has recently added support for ephemeral values and write-only arguments. There are potentially many places in the Materialize provider where one or the other of these might be appropriate to use, but for now SECRET is the most important spot for us, since we need to store lots of secrets in Materialize that come from other parts of our infra, and we don't want those to be stored in Terraform state.

User docs: https://developer.hashicorp.com/terraform/language/manage-sensitive-data/write-only
SDK docs: https://developer.hashicorp.com/terraform/plugin/sdkv2/resources/write-only-arguments

# Notes for Reviewer

* It is hard to write good tests for this as we cannot check the terraform state (as the value is not in there) and also AFAIK we cannot check the MZ instance for the secret value. So the current tests are not doing full end-to-end functional validation.
* I had some issues running the acceptance tests locally. I'm not sure if they are passing, and the CI isn't running for me due to being a first-time contributor, so I will check them once it runs.

### DIFF
--- a/docs/resources/secret.md
+++ b/docs/resources/secret.md
@@ -28,13 +28,15 @@ resource "materialize_secret" "example_secret" {
 
 ### Optional
 
+> **NOTE**: [Write-only arguments](https://developer.hashicorp.com/terraform/language/resources/ephemeral#write-only-arguments) are supported in Terraform 1.11 and later.
+
 - `comment` (String) Comment on an object in the database.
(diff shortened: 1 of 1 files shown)


bobbyiliev:
This is amazing @djahandarie! Thank you for taking the time to work on this PR! I've just tested this locally and it looks great. Just pushed a minor commit to fix the failing acceptance tests by bumping the Terraform cli in CI and added a couple of unit tests just in case. Will be merging this in a bit!

bobbyiliev:
The docs check failing is ok, it is a known problem and we have a tracking issue for this.

PR #773, opened 2025-12-05
Source versioning mysql

bobbyiliev:
As discussed, splitting the https://github.com/MaterializeInc/terraform-provider-materialize/pull/763 PR into more manageable, smaller PRs.

This PR brings over the shared base source table resources and the MySQL implementation.

The main changes to consider:
- Marking the `table` attribute as optional and deprecated for both the MySQL sources
- Introducing the new `materialize_source_table_mysql` resource which allows us to do `CREATE TABLE ... FROM SOURCE ...`

Main tracking issue: https://github.com/MaterializeInc/terraform-provider-materialize/issues/646

Fixes https://github.com/MaterializeInc/terraform-provider-materialize/issues/602

### DIFF
--- a/docs/resources/source_mysql.md
+++ b/docs/resources/source_mysql.md
@@ -52,7 +52,7 @@ resource "materialize_source_mysql" "test" {
 
 ### Optional
 
-- `all_tables` (Boolean, Deprecated) Include all tables in the source. If `table` is specified, this will be ignored.
+- `all_tables` (Boolean) **IMPORTANT**: Include all tables in the source. If `table` is specified, this will be ignored. **We strongly recommend using `materialize_source_table_mysql` resources instead** for better control and management of individual tables. See the migration guide for details.
 - `cluster_name` (String) The cluster to maintain this source.
 - `comment` (String) Comment on an object in the database.
 - `database_name` (String) The identifier for the source database in Materialize. Defaults to `MZ_DATABASE` environment variable if set or `materialize` if environment variable is not set.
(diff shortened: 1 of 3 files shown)


pkg/materialize/source_mysql.go
```diff
@@ -111,7 +117,9 @@ func (b *SourceMySQLBuilder) Create() error {
 		}
 		q.WriteString(`)`)
 	} else {
-		q.WriteString(` FOR ALL TABLES`)
+		if b.allTables {
+			q.WriteString(` FOR ALL TABLES`)
+		}
```
jubrad: Does this change the default behavior?

If someone didn't define all_tables and did not have a list of tables then upgraded to this version I think their source might be recreated with no tables?

bobbyiliev: I should have probably started with that PR: https://github.com/MaterializeInc/terraform-provider-materialize/pull/775

This should be covered in the migration guide, tldr is: use a `ignore_changes` lifecycle block to ignore changes while also defining the tables using the new source table resources.

docs/resources/source_table_mysql.md
```diff
@@ -0,0 +1,87 @@
...
+### Nested Schema for `source`
+
+Required:
+
+- `name` (String) The source name.
+
+Optional:
+
+- `database_name` (String) The source database name. Defaults to `MZ_DATABASE` environment variable if set or `materialize` if environment variable is not set.
+- `schema_name` (String) The source schema name. Defaults to `public`.
```
martykulma: Should drop database_name and maybe add a note to schema_name (e.g. that it means database).

In MySQL, schema is synonymous with database (see [glossary](https://dev.mysql.com/doc/refman/8.4/en/glossary.html)). 

MZ `CREATE SOURCE` allows users to [specify schemas](https://materialize.com/docs/sql/create-source/mysql/#with_options), but the `CREATE CONNECTION` for MySQL doesn't take a database.

PR #877, opened 2026-06-22
Add `create_if_not_exists` to` materialize_role` to adopt pre-existing roles

bobbyiliev:
Resolves [DEX-41](https://linear.app/materializeinc/issue/DEX-41/materialize-role-collides-with-externally-provisioned-ssooidc-roles-on).

Adds an optional `create_if_not_exists` flag to `materialize_role` that adopts a role already present in the catalog (e.g. one auto-provisioned by SSO/OIDC on first login) instead of failing with "role already exists", since Materialize has no `CREATE ROLE IF NOT EXISTS`. When the role exists, Terraform takes it over and converges the configured attributes; otherwise it creates the role normally.


pkg/resources/resource_role.go
```diff
@@ -175,6 +210,46 @@ func roleCreate(ctx context.Context, d *schema.ResourceData, meta interface{}) d
...
+		if err := b.AlterPassword(password); err != nil {
+			return diag.FromErr(err)
+		}
+	}
+
+	if v, ok := d.GetOk("login"); ok {
+		if err := b.AlterLogin(v.(bool)); err != nil {
+			return diag.FromErr(err)
+		}
+	}
```
jubrad: GetOk also verifies the value is not the zero value for the schema type... in this case false is the zero value, so I think we're likely to get some unexpected behavior if this is set to false in the database.

bobbyiliev: Good catch, fixed in 8a50d8b. Now reads the raw config so an explicit false is respected, and added a test for it. Thanks!

jubrad (approved):

PR #879, opened 2026-06-26
Add materialize-terraform-provider agent skill

bobbyiliev:
Agent skill for using the Materialize Terraform provider to manage Materialize resources declaratively. Linear: DEX-36.


bobbyiliev:
Moving this to the agent-skills repo, which is the home for skills: see the new PR in MaterializeInc/agent-skills. The skill was also slimmed down there: per-resource reference now points at the auto-generated provider docs instead of duplicating them.

bobbyiliev:
New PR: https://github.com/MaterializeInc/agent-skills/pull/29

PR #881, opened 2026-06-26
Add wildcard broker matching rules for Kafka PrivateLink connections

bobbyiliev:
Adding a broker_matching_rule block to materialize_connection_kafka that emits MATCHING wildcard rules inside the BROKERS clause, so dynamically discovered brokers (e.g. Confluent Cloud) can be routed per availability zone through AWS PrivateLink alongside a static bootstrap broker. Requires the enable_kafka_broker_matching_rules feature in the region. Includes builder, resource, and acceptance tests plus regenerated docs.


pkg/materialize/connection_kafka.go
```diff
@@ -251,29 +314,18 @@ func (b *ConnectionKafkaBuilder) BuildBrokersString() string {
...
-			}
+			fb.WriteString(buildAwsPrivateLinkBrokerClause(broker.PrivateLinkConnection, broker.TargetGroupPort, broker.AvailabilityZone))
 		}
 		brokersStrings = append(brokersStrings, fb.String())
 	}
+
+	for _, rule := range b.kafkaBrokerMatchingRules {
+		fb := strings.Builder{}
+		fb.WriteString(fmt.Sprintf(`MATCHING %s`, QuoteString(rule.Pattern)))
+		fb.WriteString(buildAwsPrivateLinkBrokerClause(rule.PrivateLinkConnection, rule.TargetGroupPort, rule.AvailabilityZone))
```
jubrad: do we need this conditional?

		if broker.PrivateLinkConnection.Name != "" {
			fb.WriteString(buildAwsPrivateLinkBrokerClause(broker.PrivateLinkConnection, broker.TargetGroupPort, broker.AvailabilityZone))
		}

jubrad (approved):

pkg/materialize/connection_kafka.go
bobbyiliev: `privatelink_connection` is `Required: true` on `broker_matching_rule`, so a rule always has a PrivateLink target. The guard exists on `kafka_broker` only because there it's optional (a broker can use SSH or no tunnel). Here it'd be kind of dead code, but I'll add it for consistency.

PR #887, opened 2026-07-13
Bump Go version to 1.26

bobbyiliev:
Bumps the Go version from 1.25.8 to 1.26 across go.mod, the mock server module, and the README prerequisites. The Dockerfile and CI workflows already target 1.26 (workflows derive the version from go.mod).


def- (approved):
A case for dependabot?

PR #898, opened 2026-08-03
Honor snapshot = false on Kafka sinks

bobbyiliev:
Setting `snapshot = false` on a `materialize_sink_kafka` had no effect because the value never reached the generated `CREATE SINK` statement, so the sink was created with the server default instead. The builder now emits the configured value, and `key_not_enforced` gets the same treatment since it was reading its input the same way.

### DIFF
--- a/pkg/materialize/sink_kafka.go
+++ b/pkg/materialize/sink_kafka.go
@@ -38,7 +38,7 @@ type SinkKafkaBuilder struct {
 	key                    []string
 	format                 SinkFormatSpecStruct
 	envelope               KafkaSinkEnvelopeStruct
-	snapshot               bool
+	snapshot               *bool
 	headers                string
 	keyNotEnforced         bool
 	partitionBy            string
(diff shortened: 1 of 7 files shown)


pkg/materialize/sink_kafka.go
```diff
@@ -111,8 +112,12 @@ func (b *SinkKafkaBuilder) Envelope(e KafkaSinkEnvelopeStruct) *SinkKafkaBuilder
 	return b
 }
 
+// Snapshot records the requested SNAPSHOT value. Both true and false are
+// meaningful, so the builder tracks whether it was set at all and omits the
+// option entirely otherwise.
 func (b *SinkKafkaBuilder) Snapshot(s bool) *SinkKafkaBuilder {
 	b.snapshot = s
+	b.snapshotSet = true
```
jubrad: NIT:
is snapshopSet there just to give us functionally `Option<bool>`

If so I feel like creating a generic might be cleaner if need this pattern elsewhere

```golang
type Optional[T any] struct {
    value T
    isSet bool
}
```

jubrad (approved):
Approving with one readability nit

PR #902, opened 2026-08-03
Use errors.Is/As for error checks and report failed rollback drops

bobbyiliev:
Error checks that compared or asserted concrete types directly now use `errors.Is`/`errors.As`, so a wrapped not-found error still removes the resource from state rather than failing the plan. Also reports a warning when the cleanup drop after a failed create fails, which previously left an orphaned object with no indication, and replaces two panic paths with errors.

### DIFF
--- a/pkg/materialize/generic.go
+++ b/pkg/materialize/generic.go
@@ -40,14 +40,15 @@ type Builder struct {
 }
 
 func (b *Builder) exec(statement string) error {
-	if statement[len(statement)-1:] != ";" {
+	if !strings.HasSuffix(statement, ";") {
 		statement += ";"
 	}
 
 	_, err := b.conn.Exec(statement)
 	if err != nil {
 		log.Printf("[DEBUG] error executing: %s", statement)
-		if pgErr, ok := err.(*pgconn.PgError); ok {
+		var pgErr *pgconn.PgError
+		if errors.As(err, &pgErr) {
 			msg := fmt.Sprintf("%s: %s", pgErr.Severity, pgErr.Message)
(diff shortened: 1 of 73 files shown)


pkg/materialize/generic.go
```diff
@@ -40,14 +40,19 @@ type Builder struct {
 }
 
 func (b *Builder) exec(statement string) error {
```
jubrad, line 42: **consider** — I traced every `exec` call site and I don't think this branch is reachable: each one either builds the statement with a `fmt.Sprintf` that starts with a literal (`DROP %s %s;`, `ALTER ROLE %s %s;`, …) or with a `strings.Builder` that unconditionally writes a verb before any optional clause. So there's no path that produces `""`, which also means the `statement[len(statement)-1:]` panic it replaces wasn't reachable either.

Switching that slice to `strings.HasSuffix` is the real fix and worth keeping on its own. The explicit empty check plus `TestBuilderExecEmptyStatement` is three lines and a test guarding a state the builders can't produce — I'd drop both (and the corresponding CHANGELOG clause about an empty generated statement). Your call if you'd rather keep it as a backstop for future builders.

_written by claude on behalf of @jubrad_

jubrad (approved):
Clean, well-scoped hardening pass — the `errors.Is`/`errors.As` conversion is complete (no `== sql.ErrNoRows`, no `err.(*T)`, no bare `meta.(*utils.ProviderMeta)` left in the tree), and centralising the create-rollback into `rollbackCreate` is a nice consolidation; one non-blocking note below.

## Summary

1 consideration

_written by claude on behalf of @jubrad_

PR #913, opened 2026-08-27
Warn when MZ_HOST silently selects self-hosted mode

bobbyiliev:
A customer hit a confusing `invalid password` error after upgrading, because `MZ_HOST` was left over in their shell. Setting `host` is what selects self-hosted mode, and `host` falls back to `MZ_HOST`, so the provider switched away from Materialize Cloud even though their configuration never mentioned a host. The provider now warns when the host it used did not come from the configuration, and the docs say that `host` selects the mode. Also fixes the documented variable name for `username`, which is `MZ_USERNAME` and not `MZ_USER`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Changelog
 
+## 0.11.8 - Unreleased
+
+### Bug Fixes
+
+* The provider now warns when it selects self-hosted mode because `MZ_HOST` is set in the environment rather than `host` being set in the configuration. A leftover `MZ_HOST` silently switched the provider away from Materialize Cloud, which authenticated as the `username` default instead of the app password owner and produced a confusing `invalid password` error [#913](https://github.com/MaterializeInc/terraform-provider-materialize/pull/913).
+* Corrected the documented environment variable for `username`, which is `MZ_USERNAME` and not `MZ_USER`, and documented that `host` selects self-hosted mode [#913](https://github.com/MaterializeInc/terraform-provider-materialize/pull/913).
+
 ## 0.11.7 - 2026-08-24
 
 ### Bug Fixes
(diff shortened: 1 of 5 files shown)


jubrad (commented):
I'm not sure I like this, I think some of the assumptions around using app password to determine mode are just bad. 

Could we just stop allowing `MZ_HOST` and require users to set host in the provider config directly?

bobbyiliev:
> I'm not sure I like this, I think some of the assumptions around using app password to determine mode are just bad.
> 
> Could we just stop allowing `MZ_HOST` and require users to set host in the provider config directly?

@jubrad yea I agree that this one here is not ideal but two things to keep in mind:

- First a small correction: the app password does not decide the mode. The mode is still only `host != ""`, same as before. The `mzp_` check is inside that branch and only decides if we show the warning or not. But you are right that it is a weak guess. I am happy to remove it.

- Second, I do not think removing `MZ_HOST` really solves the problem. I tested it. If the host must be in the config, then the migration for users is `host = var.mz_host` with `TF_VAR_mz_host`. When that variable is not set, the provider silently switches to Cloud and tries to create a Frontegg client. The host was in the config the whole time, only the value came from the environment again. So we would break the setup that our docs recommend, and we would get the same bug, just in the other direction.

So I think the environment variable is not really the problem. The problem is that we decide the backend based on whether a string is empty or not. That stays true if the value comes from `MZ_HOST` or from a `TF_VAR_` that nobody set.

An explicit `mode` would cover both directions. With `mode = "cloud|saas|better-name-here"` we ignore any host that is set, so such cases cannot happen. With `mode = "direct"` we return a clear error if the host is missing, instead of switching to Cloud without telling anyone. And if someone sets both, we can reject it.

It is also the smaller change for users. Adding a new argument breaks nobody, while removing `MZ_HOST` breaks everyone who follows our current docs.

We can also do both if you prefer, and deprecate `MZ_HOST` after `mode` is in. But if we only do one thing, I would pick `mode`.

PR #917, opened 2026-09-09
Re-resolve a cluster grant's id when the cluster was swapped or renamed

jazz-mo:
`materialize_cluster_grant` addresses its cluster by name but stores the cluster's catalog id in the resource id. `ALTER CLUSTER ... SWAP WITH` exchanges names, so the canonical name ends up on a different cluster and the stored id refers to a dropped one. `grantRead` then gets no rows, drops the resource from state, and the next plan wants to create a grant that already exists.

The read now re-resolves the id from `cluster_name` when the stored id is gone. A cluster that exists under neither the id nor the name is still removed from state.

Also carries the fix from #918, which is a prerequisite rather than a nice to have. `grantRead` cleared the id when the privilege was missing but set it again before returning, so nothing ever left state that way. Without it, re-resolving to a cluster that does not carry the grant reports no drift at all, which is worse than the bug this PR fixes. Verified locally against a real Materialize: a swap with grants copied plans clean, a swap without them plans the grant back in.

### DIFF
--- a/pkg/resources/resource_grant.go
+++ b/pkg/resources/resource_grant.go
@@ -51,6 +51,20 @@ func grantRead(ctx context.Context, d *schema.ResourceData, meta interface{}) di
 	}
 
 	p, err := materialize.ScanPrivileges(metaDb, materialize.EntityType(key.objectType), key.objectId)
+	if errors.Is(err, sql.ErrNoRows) && materialize.EntityType(key.objectType) == materialize.Cluster {
+		// A swapped or renamed cluster leaves a dropped id in state, re-resolve it from the configured name
+		c, clusterErr := materialize.ScanCluster(metaDb, d.Get("cluster_name").(string), true)
+		if clusterErr != nil && !errors.Is(clusterErr, sql.ErrNoRows) {
+			return diag.FromErr(clusterErr)
+		}
+		if clusterErr == nil {
+			key.objectId = c.ClusterId.String
+			ie := strings.Split(i, "|")
+			ie[2] = key.objectId
+			i = strings.Join(ie, "|")
+			p, err = materialize.ScanPrivileges(metaDb, materialize.Cluster, key.objectId)
+		}
+	}
 	if errors.Is(err, sql.ErrNoRows) {
 		log.Printf("[WARN] grant (%s) not found, removing from state file", d.Id())
 		d.SetId("")
--- a/pkg/resources/resource_grant.go
+++ b/pkg/resources/resource_grant.go
@@ -87,6 +87,7 @@ func grantRead(ctx context.Context, d *schema.ResourceData, meta interface{}) di
 		log.Printf("[DEBUG] %s object does not contain privilege %s", i, privilege)
 		// Remove id from state
 		d.SetId("")
+		return nil
 	}
 
 	d.SetId(utils.TransformIdWithRegion(string(region), i))
--- a/pkg/resources/resource_grant.go
+++ b/pkg/resources/resource_grant.go
@@ -52,17 +52,22 @@ func grantRead(ctx context.Context, d *schema.ResourceData, meta interface{}) di
 
 	p, err := materialize.ScanPrivileges(metaDb, materialize.EntityType(key.objectType), key.objectId)
 	if errors.Is(err, sql.ErrNoRows) && materialize.EntityType(key.objectType) == materialize.Cluster {
-		// A swapped or renamed cluster leaves a dropped id in state, re-resolve it from the configured name
-		c, clusterErr := materialize.ScanCluster(metaDb, d.Get("cluster_name").(string), true)
-		if clusterErr != nil && !errors.Is(clusterErr, sql.ErrNoRows) {
-			return diag.FromErr(clusterErr)
-		}
-		if clusterErr == nil {
-			key.objectId = c.ClusterId.String
-			ie := strings.Split(i, "|")
-			ie[2] = key.objectId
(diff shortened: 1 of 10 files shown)


bobbyiliev (commented):
Thanks for this, the fix looks right and the two tests cover the happy path and the case where the cluster is gone for good. Left a few comments inline.

If it is easier I am happy to push the changes to your branch and take it from here, just say the word.

pkg/resources/resource_grant.go
```diff
@@ -51,6 +51,20 @@ func grantRead(ctx context.Context, d *schema.ResourceData, meta interface{}) di
 	}
 
 	p, err := materialize.ScanPrivileges(metaDb, materialize.EntityType(key.objectType), key.objectId)
+	if errors.Is(err, sql.ErrNoRows) && materialize.EntityType(key.objectType) == materialize.Cluster {
+		// A swapped or renamed cluster leaves a dropped id in state, re-resolve it from the configured name
+		c, clusterErr := materialize.ScanCluster(metaDb, d.Get("cluster_name").(string), true)
```
bobbyiliev: `d.Get("cluster_name").(string)` panics when the field is not in the schema, since Get returns nil there. All ten grant resources share `grantRead` and use `ImportStatePassthroughContext`, so importing a `...|CLUSTER|...` id into something like `materialize_database_grant` crashes the provider instead of erroring.

I tried it on your branch and it panics with `interface conversion: interface {} is nil, not string`. A checked assertion sorts it.

pkg/resources/resource_grant.go
```diff
@@ -51,6 +51,20 @@ func grantRead(ctx context.Context, d *schema.ResourceData, meta interface{}) di
 	}
 
 	p, err := materialize.ScanPrivileges(metaDb, materialize.EntityType(key.objectType), key.objectId)
+	if errors.Is(err, sql.ErrNoRows) && materialize.EntityType(key.objectType) == materialize.Cluster {
```
bobbyiliev, line 54: `materialize_cluster` makes name based identity opt in with `identify_by_name`, which defaults to false. Here cluster grants always fall back to the name.

Should it be opt in too? Quietly pointing the resource at a different cluster feels like something people would want to choose. Real question, not a nit.

pkg/resources/resource_grant.go
```diff
@@ -51,6 +51,20 @@ func grantRead(ctx context.Context, d *schema.ResourceData, meta interface{}) di
...
+		c, clusterErr := materialize.ScanCluster(metaDb, d.Get("cluster_name").(string), true)
+		if clusterErr != nil && !errors.Is(clusterErr, sql.ErrNoRows) {
+			return diag.FromErr(clusterErr)
+		}
+		if clusterErr == nil {
+			key.objectId = c.ClusterId.String
+			ie := strings.Split(i, "|")
+			ie[2] = key.objectId
+			i = strings.Join(ie, "|")
+			p, err = materialize.ScanPrivileges(metaDb, materialize.Cluster, key.objectId)
```
bobbyiliev: Heads up, the `d.SetId("")` further down on line 84 is overwritten by line 87 with no return in between, so that path never removes anything from state. Old bug on main and not yours, I opened #918 for it.

It matters here though. Once we re-resolve to a different cluster, if that cluster does not carry the grant we keep the resource in state and the drift is invisible.

pkg/resources/resource_grant_test.go
```diff
@@ -42,3 +42,74 @@ func TestResourceGrantPrivilegeReadIdMigration(t *testing.T) {
...
+
+		if err := grantRead(context.TODO(), d, db); err != nil {
+			t.Fatal(err)
+		}
+
+		if d.Id() != "" {
+			t.Fatalf("unexpected id of %s", d.Id())
+		}
+	})
+}
```
bobbyiliev, line 115: Once #918 is in, could we add a third case here where the new cluster exists but does not have the grant, asserting it leaves state? That is the one I would most want covered.

pkg/testhelpers/mock_scans.go
```diff
@@ -119,11 +119,20 @@ func MockClusterScan(mock sqlmock.Sqlmock, predicate string) {
...
 	ir := mock.NewRows([]string{"id", "name", "managed", "size", "replication_factor", "disk", "availability_zones", "comment", "owner_name", "privileges"}).
 		AddRow("u1", "cluster", true, "small", 2, true, az, "comment", "joe", defaultPrivilege)
-	mock.ExpectQuery(q).WillReturnRows(ir)
+	mock.ExpectQuery(clusterScanQuery(predicate)).WillReturnRows(ir)
+}
+
+// MockClusterScanNoRows mocks a cluster scan that matches nothing, simulating a
+// cluster that no longer exists.
+func MockClusterScanNoRows(mock sqlmock.Sqlmock, predicate string) {
+	mock.ExpectQuery(clusterScanQuery(predicate)).WillReturnRows(mock.NewRows([]string{"id"}))
```
bobbyiliev: Builds its rows with just the `id` column. Works because there are no rows, but listing the same columns as `MockClusterScan` would be less surprising later.

jazz-mo:
> Thanks for this, the fix looks right and the two tests cover the happy path and the case where the cluster is gone for good. Left a few comments inline.
> 
> If it is easier I am happy to push the changes to your branch and take it from here, just say the word.

If you could please take it from here, I'd appreciate it. I am not fully familiar with the codebase, so am happy to let you take the driver's seat!

bobbyiliev (approved):

PR #918, opened 2026-09-09
Remove a grant from state when the privilege is no longer present

bobbyiliev:
`grantRead` cleared the id when the privilege was gone from the object, but then set it again two lines later before returning, so a revoked grant stayed in state and the next plan saw no drift.

Adding the missing return exposed that the PUBLIC grant tests were never checking the privilege at all, since the shared mock privileges had no PUBLIC row, so that fixture gets one too.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+### Bug Fixes
+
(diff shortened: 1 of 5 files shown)


bobbyiliev:
Folded into #917. It turned out to be a prerequisite for that fix rather than an independent cleanup, so keeping them apart risked merging them in the wrong order and shipping silent drift.

